PR #49, opened 2022-03-07
Print exception on failure in certain cases

banool:
I'm running into an error with remote database destroy and trying to debug it. I see that while other errors get printed, this one doesn't. So I'm adding logging for that, as well as unifying some of the language and the log level.


vrtmrz:
Thank you for your contribution!
LocalPouchDB is the most complicated area of this plugin! 
It will be a very good improvement!

PR #89, opened 2022-07-12
Removing the step to enable history from the README

MohamedBassem:
While following the README, it told me to enable "Use History" though I couldn't find this setting anywhere. Turned out, it became the default as of https://github.com/vrtmrz/obsidian-livesync/commit/89de551fd77f695dae3c0e514763c116345f18cd with no way to be configured. So this PR, removes this step from the readme.


vrtmrz:
Thank you for your Pull-request!
Merged!

PR #98, opened 2022-08-04
improved grammar & fixed typos in docs and plugin texts

Miigon:
I am not a native speaker, but still I think this should improve the legibility of documents and texts inside the plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,32 +3,32 @@
 [Japanese docs](./README_ja.md).
 
 Self-hosted LiveSync is a community implemented synchronization plugin.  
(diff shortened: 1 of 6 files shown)


vrtmrz:
I'm really appriciating for your pull-request! 
I will merge them after checking the implications.

Miigon:
Thanks! There was some part in readme that I could not fully understand (mainly technical details that I'm not sure about how correct my assumption is). I've left some comments there and maybe after further clarification we can make it more explicit.

vrtmrz (approved):
Thank you for the great improvement!
I'll make a few amendments.

PR #101, opened 2022-08-09
Add Chinese documentations

Miigon:
This plugin is so cool I wanted to share it with more people, so I added Chinese translation for setup_own_server.md and README.md (didn't translate the IBM Cloudant guide since it's not very popular here, and those who actually uses it would know how to read English anyway.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Self-hosted LiveSync
 
-[Japanese docs](./README_ja.md).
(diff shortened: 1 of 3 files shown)


vrtmrz:
Wow! Thank you for your contribution! I'm really honoured.
Actually, I don't know Chinese very well, but I think your contribution will be fine.

PR #114, opened 2022-09-12
Fixed Docker command in docs.

JEndler:
Hey, I've noticed that running the docker command  
```
$ docker run --rm -it -e COUCHDB_USER=admin -e COUCHDB_PASSWORD=password -v .local.ini:/opt/couchdb/etc/local.ini -p 5984:5984 couchdb
```
runs into the following error:

```
docker: Error response from daemon: create .local.ini: ".local.ini" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```

This is because **-v** expects full paths as documented [here](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems).

I fixed this by adding this to the docs:

```
$ docker run --rm -it -e COUCHDB_USER=admin -e COUCHDB_PASSWORD=password -v /path/to/local.ini:/opt/couchdb/etc/local.ini -p 5984:5984 couchdb
*Remember to replace the path with the path to your local.ini*
```

Thanks for developing such a great tool 👍🏼

### DIFF
--- a/docs/setup_own_server.md
+++ b/docs/setup_own_server.md
@@ -32,16 +32,18 @@ max_age = 3600
 
 Make `local.ini` and run with docker run like this, you can launch the CouchDB.
 ```
-$ docker run --rm -it -e COUCHDB_USER=admin -e COUCHDB_PASSWORD=password -v .local.ini:/opt/couchdb/etc/local.ini -p 5984:5984 couchdb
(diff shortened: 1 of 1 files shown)


vrtmrz:
Thank you for your pull request!
I wonder why I didn't notice that the relative path cannot be used.

Merged!

> Thanks for developing such a great tool 👍🏼
Now you too! Thank you again!

PR #167, opened 2023-01-28
WIP: Improved Skip Patterns

eth-p:
**This is currently a work in progress commit.**
I'm making a draft pull request to get your feedback on the feature and its implementation.

The idea of this pull request to improve the user experience of the "Skip patterns" setting by providing more types of patterns, a better user interface for editing the patterns, and visual feedback to know if patterns are working.

image

**New Features:**
- Custom settings UI for adding/removing/changing the skip patterns.
  - Provides feedback for patterns:
     ❌ --> Invalid pattern
     ✅ --> Valid pattern
- Glob-style patterns for skip patterns.
  - `*` matches everything except directories
  - `**` matches directories
  - `{a,b}` matches either `a` or `b`
    - Like bash, `{a}` is invalid, so it literally matches `{a}`.

### Done
- Add support for glob-style skip patterns.
- Add setting component for adding and removing skip patterns.

### To-Do
- Show a different icon if the pattern does not match any files in the vault (⚠️, maybe? or a non-green ✔)
- Test that the regex created by the glob pattern compiler works as expected.

### DIFF
--- a/src/PatternSet.ts
+++ b/src/PatternSet.ts
@@ -1,0 +1,473 @@
+import { Setting } from "obsidian";
+
+// Pattern Grammar:
+type RegexPattern = { regex: string };
+type GlobPattern = { glob: string };
+
+type ValidPattern = RegexPattern | GlobPattern;
+type InvalidPattern = { invalid: ValidPattern }
+
+export type Pattern = ValidPattern | InvalidPattern;
+
+// Constants:
+const REGEX_ESCAPE_METACHARS = /[.*+?^${}()|[\]\\]/g;
+const PATH_SEPARATOR = '/';
+
+/**
+ * An ordered set of file-matching patterns.
+ */
+export class PatternSet {
+    private rawPatterns: Pattern[];
+    private compiledPatterns: (RegExp|InvalidPattern)[];
+
+    public constructor(rawPatterns: Pattern[]) {
+        this.rawPatterns = rawPatterns.slice();
+        this.compiledPatterns = rawPatterns.map(compilePattern);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


vrtmrz:
Thank you for proposing great new features.
I think it would be very useful! 

I have just seen your codes only on mobile yet, this is based on only this note, and I hope to see the details later!

> Show a different icon if the pattern does not match any files in the vault (⚠️, maybe? or a non-green ✔)

If we count the number of matched documents, how about indicating that? 
Like: ✅ (2), ✅(0), ❌(-)
And, It might be also nice to display the beginning of the list as a tooltip of the number.

src/PatternSet.ts
```diff
@@ -0,0 +1,473 @@
...
+    //   '{a,b}' => matches either 'a' or 'b'
+    if ('glob' in pattern) {
+        const globAsRegex = pattern.glob
+            .replace(REGEX_ESCAPE_METACHARS, "\\$&")  // Escape regex metachars.
+            .replace("\\*", `[^${PATH_SEPARATOR}]+`)  // Glob: '*'
+            .replace("\\*\\*", ".*")                             // Glob: '**'
+            .replace(/\\\{([^}]+(?:,[^}]+)+)\\\}/,               // Glob: '{a,b}'
+                (_, [arg]) => `(?:${arg.split(",").join("|")})`)
+
+        return new RegExp(globAsRegex);
```
pwall2222, line 171: Translating a glob, to RegExp, means that things like `!` can't work. 
Using something like `picomatch` or `micromatch` would be better in my opinion.

pwall2222:
Also, is this PR dead?

PR #170, opened 2023-02-03
add tip : redflag.md can be a directory which is easier to create on iOS

hilsonp:


vrtmrz (approved):
Thank you for this tip! It might be so useful. 

This has led me to write some more tips!

PR #222, opened 2023-06-01
Added troubleshooting guide

Hugo-Persson:
Added guide for troubleshooting. This was an easy I was having and this is the way I solved it. Some improvements might be good but I think it is a good start to collect ways to troubleshoot and fix problems.


vrtmrz:
Thank you for your pull request! 
I think that it will be a basis for the future.

PR #224, opened 2023-06-05
[Feature] Add password protection to askString function

antoKeinanen:
Adds option to create hidden password inputs using `askString`. This is implemented using the `-webkit-text-security` CSS style and replaces all text in the input field with disc (●) characters. The change is only visual. This change only affects setting import URI passwords as I found nothing else that would need to be hidden.


vrtmrz (approved):
Thank you for the PR! 
That field should be masked!

PR #236, opened 2023-07-03
CouchDB config corrections, reverse proxy documentation, other additions

bioluks:
- [enable_cors is now under section [chttpd] since version 3.2](https://docs.couchdb.org/en/latest/config/http.html#chttpd/enable_cors).
- Not everyone will use version 3.2+ so I made it clear that's a new thing.
- Added `bind_address` for making sure the container uses all interfaces given; I myself use Traefik as a reverse proxy and using this somehow made it work, since this option does not harm other reverse proxies and there are lots of Traefik users this option should be there for wider compatibility.
- Added spaces between `origins` elements and removed spaces between the `methods` elements because it's like this in the official documentation, see [1](https://docs.couchdb.org/en/latest/config/http.html#cors/origins) and [2](https://docs.couchdb.org/en/latest/config/http.html#cors/methods).
- Added a write permission warning since many newbies had this mistake with couchdb, including myself.

### DIFF
--- a/docker-compose.traefik.yml
+++ b/docker-compose.traefik.yml
@@ -1,0 +1,46 @@
+# For details and other explanations about this file refer to:
+# https://github.com/vrtmrz/obsidian-livesync/blob/main/docs/setup_own_server.md#traefik
+
+version: "2.1"
+services:
+  couchdb:
+    image: couchdb:latest
+    container_name: obsidian-livesync
+    user: 1000:1000
+    environment:
+      - COUCHDB_USER=username
+      - COUCHDB_PASSWORD=password
+    volumes:
+      - ./data:/opt/couchdb/data
+      - ./local.ini:/opt/couchdb/etc/local.ini
+    # Ports not needed when already passed to Traefik
+    #ports:
+    #  - 5984:5984
+    restart: unless-stopped
+    networks:
+      - proxy
+    labels:
+      - "traefik.enable=true"
+      # The Traefik Network
+      - "traefik.docker.network=proxy"
+      # Don't forget to replace 'obsidian-livesync.example.org' with your own domain
+      - "traefik.http.routers.obsidian-livesync.rule=Host(`obsidian-livesync.example.org`)"
+      # The 'websecure' entryPoint is basically your HTTPS entrypoint. Check the next code snippet if you are encountering problems only; you probably have a working traefik configuration if this is not your first container you are reverse proxying.
+      - "traefik.http.routers.obsidian-livesync.entrypoints=websecure"
+      - "traefik.http.routers.obsidian-livesync.service=obsidian-livesync"
+      - "traefik.http.services.obsidian-livesync.loadbalancer.server.port=5984"
+      - "traefik.http.routers.obsidian-livesync.tls=true"
+      # Replace the string 'letsencrypt' with your own certificate resolver
+      - "traefik.http.routers.obsidian-livesync.tls.certresolver=letsencrypt"
+      - "traefik.http.routers.obsidian-livesync.middlewares=obsidiancors"
+      # The part needed for CORS to work on Traefik 2.x starts here
+      - "traefik.http.middlewares.obsidiancors.headers.accesscontrolallowmethods=GET,PUT,POST,HEAD,DELETE"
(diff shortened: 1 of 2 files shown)


bioluks:
Hi, wasn't able to find documentation on how to selfhost couchdb behind traefik, because of this I overexplained a bit. Shouldn't harm anybody.

The second commit I added the Traefik reverse proxy files and documentation, along with a Table of Contents.

docker-compose.traefik.yml
```diff
@@ -0,0 +1,46 @@
...
+      # The 'websecure' entryPoint is basically your HTTPS entrypoint. Check the next code snippet if you are encountering problems only; you probably have a working traefik configuration if this is not your first container you are reverse proxying.
+      - "traefik.http.routers.obsidian-livesync.entrypoints=websecure"
+      - "traefik.http.routers.obsidian-livesync.service=obsidian-livesync"
+      - "traefik.http.services.obsidian-livesync.loadbalancer.server.port=5984"
+      - "traefik.http.routers.obsidian-livesync.tls=true"
+      # Replace the string 'letsencrypt' with your own certificate resolver
+      - "traefik.http.routers.obsidian-livesync.tls.certresolver=letsencrypt"
+      - "traefik.http.routers.obsidian-livesync.middlewares=obsidiancors"
+      # The part needed for CORS to work on Traefik 2.x starts here
+      - "traefik.http.middlewares.obsidiancors.headers.accesscontrolallowmethods=GET,PUT,POST,HEAD,DELETE"
```
vrtmrz, line 37: Does the `headers` middleware actually need? CORS is expected to be handled by CouchDB.
Could you please check the behaviour without this middleware?

docs/setup_own_server.md
```diff
@@ -127,6 +140,77 @@ Set the A record of your domain to point to your server, and host reverse proxy
...
+      - "traefik.http.routers.obsidian-livesync.rule=Host(`obsidian-livesync.example.org`)"
+      # The 'websecure' entryPoint is basically your HTTPS entrypoint. Check the next code snippet if you are encountering problems only; you probably have a working traefik configuration if this is not your first container you are reverse proxying.
+      - "traefik.http.routers.obsidian-livesync.entrypoints=websecure"
+      - "traefik.http.routers.obsidian-livesync.service=obsidian-livesync"
+      - "traefik.http.services.obsidian-livesync.loadbalancer.server.port=5984"
+      - "traefik.http.routers.obsidian-livesync.tls=true"
+      # Replace the string 'letsencrypt' with your own certificate resolver
+      - "traefik.http.routers.obsidian-livesync.tls.certresolver=letsencrypt"
+      - "traefik.http.routers.obsidian-livesync.middlewares=obsidiancors"
+      # The part needed for CORS to work on Traefik 2.x starts here
```
vrtmrz, line 187: If not required, please address also here.

vrtmrz (changes requested):
Thank you for the excellent PR! I have a few points of concern that I would be most grateful if you could kindly address. May I ask you to check the operation without middleware?

All others look very refined! Thank you for your efforts that is much appreciated.

docker-compose.traefik.yml
bioluks: To my knowledge and experience it is **definitely** required, a default Traefik config won't work without these headers allowed [explicitly](https://doc.traefik.io/traefik/middlewares/http/headers/#cors-headers). I tested it with my production and local test server. Someone can also define these 'middlewares' inside other files too and [reference it](https://doc.traefik.io/traefik/providers/overview/#provider-namespace) inside Traefik's docker-compose.yml file, but for the sake of simplicity having all the Traefik labels in the docker-compose.yml file is better for getting the job done faster; generally working with one file in this case.

It may look weird at first that we have to define middlewares for allowing CORS and AccessControlAllowCredentials headers to be passed, but this is the Traefik way and it works reliably.

vrtmrz: Thank you for the detail! I have understood the behaviour of traefik. And I would love to apologise for my misunderstanding and withdraw this point.

I have not understood traefik on Docker well. This will enrich our knowledge!
Everything has got clear. Thank you very much!

vrtmrz (approved):

PR #237, opened 2023-07-04
Partially corrected CORS checks on the client side

bioluks:
Hi, like explained [here](https://github.com/vrtmrz/obsidian-livesync/pull/236) since version 3.2 the `enable_cors` option is under `[chttpd]` and not `[httpd]`. I only moved the `enable_cors` option. Result on my testing rig:

cors-check-fix

\
 The `cors.origins` check still needs an addition, it works just fine when there are no spaces between the origins, on my setup I have them so that's the reason why I have this error; since both are valid in the local.ini configuration both can be accepted? Please check code comments.

### DIFF
--- a/src/ObsidianLiveSyncSettingTab.ts
+++ b/src/ObsidianLiveSyncSettingTab.ts
@@ -372,6 +372,12 @@ export class ObsidianLiveSyncSettingTab extends PluginSettingTab {
                                 } else {
                                     addResult("✔ chttpd_auth.require_valid_user is ok.");
                                 }
+                                if (responseConfig?.chttpd?.enable_cors != "true") {
+                                    addResult("❗ chttpd.enable_cors is wrong");
+                                    addConfigFixButton("Set chttpd.enable_cors", "chttpd/enable_cors", "true");
+                                } else {
+                                    addResult("✔ chttpd.enable_cors is ok.");
+                                }
                                 // HTTPD check
                                 //  Check Authentication header
                                 if (!responseConfig?.httpd["WWW-Authenticate"]) {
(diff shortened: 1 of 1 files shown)


bioluks:
Also a couchdb version check can be added; if it's lower than 3.2 `[httpd].enable_cors` would be the right path

vrtmrz:
Sorry for my absence! and thank you for making this PR! I did not realise it yet!
This steady fixing would help us! However, as you pointed out, there might be users of v3.2 before.
Would you mind if I ask you to leave the previous check logic as a warning with a comment like `(v3.1 or before)`? 
(i.e., `⚠ chttpd.enable_cors is wrong (v3.1 or before)` ).

It would be harmless even if `chttpd.enable_cors` is configured after v3.2.

bioluks:
No thing, thanks for this great project!

> Would you mind if I ask you to leave the previous check logic as a warning with a comment like (v3.1 or before)?
(i.e., ⚠ chttpd.enable_cors is wrong (v3.1 or before) ).

We can do that, but wouldn't it be better if we do a server version check first? Visiting the server gives a JSON response like the following already:

responsejson

So what I'm saying is if it's lower than 3.2 it should say `httpd.enable_cors is ok` and shouldn't print information about `chttpd.enable_cors`; but if it's 3.2 and higher it should say `chttpd.enable_cors is ok` without the unneccesary `httpd.enable_cors` part in this case. The `https.enable_cors is wrong` warning made me lose hope first time setting up couchdb. Short comments next to important errors are not working so well for the majority, my personal thought. What do you think?

vrtmrz:
Yes, checking the remote version should be the correct way. As your mentioned, we should make an easy-usable fix than an easy fix. 
Thank you for bringing me back to the attitude we should be.

Would you mind if I ask you to add that logic as well, please?

cwsqer006:
DFGTYHKL;

cwsqer006:
DFGHJKL/

PR #244, opened 2023-07-14
fix: separate languages

samyarkd:
when i first saw it i thought it's one connected sentence so, i thought it would be nice to separate them using a dash.


vrtmrz (approved):
Thank you very much! It has been clarified.

vrtmrz:
Thank you very much! It has been clarified.

PR #276, opened 2023-09-01
Add Quick setup Chinese translation.

cgcel:


vrtmrz:
Sorry for the late response. And, thank you for your brilliant improvement! 
I could not understand Chinese, however, it looks fine when looking through some translation engines. 
It has been merged!

PR #336, opened 2023-11-24
Lower payload size limit and batch limit from 10 to 2 

toon159:
I want to lower the payload size limit and batch limit from 10 to 2, because I think they are too high.
This might solve the “Replication error” problem, since the IBM Cloudant lite plan has a 20 reads/sec and 10 writes/sec limit that is really easy to exceed.
I haven't tested it yet, but this is my idea.


cwsqer006:
hi

vrtmrz:
Thank you for opening this PR! This PR is very clear and works for specific scenes. So I will be happy to merge it!

However, if we set the lower batch size, requests will be performed more frequently. Since v0.22.0, the request interval is more solidly kept. Would you mind if I ask you to check that behaviour, please? 
Unfortunately, if it does not work well, I am very sad to ask you this but could you please report it as the issue? You are always welcome to report back! Thank you so much for your contribution!

PR #389, opened 2024-03-17
Fixed docker-compose command in docs

Seeker0472:
Hey,there! 
I've noticed that running the docker command 
```
docker compose up -d
```
resulted into the following error
```
unknown shorthand flag: 'd' in -d
See 'docker --help'.
```
It is because the `docker-compose` command was mistakenly written as `docker compose`

I've fixed this by changing  "docker compose" into "docker-compose".

Thank you for reviewing this pull request,have a good time!

### DIFF
--- a/docs/setup_own_server_cn.md
+++ b/docs/setup_own_server_cn.md
@@ -91,7 +91,7 @@ services:
 最后, 创建并启动容器:
 ```
 # -d will launch detached so the container runs in background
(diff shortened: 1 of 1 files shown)


vrtmrz:
Thank you for your PR! I am so happy to merge this!
In fact, the Chinese version had been a little behind the English version, but I could not fix it. Your contribution has made it better!

PR #479, opened 2024-08-07
add username and password to setup URI instructions

fuhrysteve:


Joao-1:
Although it seems meaningless, I think it can help to understand the process without having to go into the code to know all the variables.

vrtmrz:
I am sorry for missing this! 

Indeed, it should be so important! Without doing step 1, these variables should be missing.

I am so happy to merge this! Thank you for your contribution!

PR #533, opened 2024-11-11
chore(formatting): adds prettier for consistent coding style

doublethefish:
Adds auto-formating tools for better code-clarity and collaboration.

## Why this patch?
The plugin's code is quite complex and missing a lot of dev-ops types
checks and standards. One of the key problems with codebases like this is
on-boarding new developers to the codebase (like myself). When there is
a consistent and enforced coding-style (irrespective of what the coding
style is) it makes it _significantly_ easier for collaborators and
maintainers to get on with the job in hand. It also, from a day-2-day
developer perspective, significantly reduces cognitive overhead re
reading code.
Finally this is a "trial balloon" PR, if this patch is accepted I will
likely do more work on testing and docs for the project.

## Notes on the config
- The new prettier config is a non-standard setup, but a close-match to
  how the code _currently_ looks.
- 120 col-width print width (instead of the better and more
  information-dense 88), this is so the diff after applying prettier to
  the code is less disruptive, whilst still showing the benefits of using
  a prettier.
- We use `tabWidth` setting of 4 as the code uses that more common
  setting instead of the more compact 2 spaces - note that 2 often leads
  to more readable and compact code.
- We enforce trailing commas, as that seems to be the norm in this
  code-base. We choose the `es5` standard here.
- We enforce tailing semi-colons (`semi`) as the majority of code used
  that flavour of `js`/`ts`.
- For now we only run on code and not json files.

## Notes on the tooling
This is designed such that `npm run pretty` re-formats the code for
development, and when integrated with ci/cd, `prettyCheck` will return
non-zero exit codes when formatting doesn't match the coding standards.


vrtmrz:
Thank you for your PR! I think this is an excellent workspace setup! It must be quite helpful for us.
However, could you refine it further with a bit of additional work? Here are some suggestions that I noticed:

- Please ensure it is cross-platform runnable. (Windows cannot accept single quotes in filenames in the scripts.)
- Additionally, I noticed that it cannot pass extra arguments (such as --check) to the subscripts. (However, This could be attributable to my setup mistakes).
- Please add main_org.js to the .prettierignore.

I know it is a little rude to ask for more of your kind contributions, but please help us make it the best it can be!

doublethefish:
> * Please ensure it is cross-platform runnable. (Windows cannot accept single quotes in filenames in the scripts.)

👍 Done with caveats: I don't have access to a windows box at the moment but I think the escaped double quotes `\"` should work.

> * Additionally, I noticed that it cannot pass extra arguments (such as --check) to the subscripts. (However, This could be attributable to my setup mistakes).

👍 My oversight was using `yarn` invocations instead of using `npm`. `npm` requires jumping through a couple more hoops. Thank you for highlighting this! 🙏

> * Please add main_org.js to the .prettierignore.

👍

> I know it is a little rude to ask for more of your kind contributions, but please help us make it the best it can be!

全く失礼ではありません。迅速なご返信に感謝いたします。

Also note that I had mistakenly not re-run the prettier on the original code, after changing the printWidth to `120`; doing so makes this initial diff much smaller and easier to reason about.

vrtmrz:
I have confirmed your kindly modification, and working fine also on Windows machine! 
Thank you for your contribution! I am so happy to merge it!

P.S. And, very gratifyingly, my knowledge also has improved. I appreciate so much for you and so much for the Engineers' power of improving the world by sharing good things. そして日本語でのコメントもありがとうございます。ほっとします！

PR #537, opened 2024-11-12
chore(cicd): add workflow to run checks on push to main and pull-reqs

doublethefish:
This configures the submodule and runs `npm run check` on the code.

This is a first-pass of getting the checks working before I make any other changes, and help me ensure I don't break the project.


doublethefish:
For the workflow checks to pass we will require https://github.com/vrtmrz/livesync-commonlib/pull/7 to be accepted upstream, and the git-submodule updated with that change.

PR #543, opened 2024-11-18
Fix up wording of messages

Volkor3-16:
Hi vorotamoroz, thanks for making this plugin!

I reworded a bunch of messages and UI elements, they caused me quite a bit of confusion as I was setting this up!

I've never done any typescript before, so I could have broken everything (there were a lot of errors in vscode, but I /think/ that's down to me not having everything setup right).

When you have the time, please check over this to make sure I didn't break everything - I didn't test any of these changes (no idea how :disappointed:)
Happy to make further changes if you have any issues :smiley: 

Cheers

### DIFF
--- a/src/modules/features/ModuleSetupObsidian.ts
+++ b/src/modules/features/ModuleSetupObsidian.ts
@@ -139,11 +139,11 @@ export class ModuleSetupObsidian extends AbstractObsidianModule implements IObsi
                     newSettingW.encryptedPassphrase = "";
                     newSettingW.encryptedCouchDBConnection = "";
                     newSettingW.additionalSuffixOfDatabaseName = `${"appId" in this.app ? this.app.appId : ""}`;
-                    const setupJustImport = "Just import setting";
-                    const setupAsNew = "Set it up as secondary or subsequent device";
-                    const setupAsMerge = "Secondary device but try keeping local changes";
-                    const setupAgain = "Reconfigure and reconstitute the data";
-                    const setupManually = "Leave everything to me";
+                    const setupJustImport = "Don't sync anything, just apply the settings.";
+                    const setupAsNew = "This is a new client - sync everything from the remote server.";
+                    const setupAsMerge = "This is an existing client - merge existing files with the server.";
+                    const setupAgain = "This is a new server - reset remote server and use local files.";
+                    const setupManually = "Continue and configure manually.";
                     newSettingW.syncInternalFiles = false;
                     newSettingW.usePluginSync = false;
                     newSettingW.isConfigured = true;
@@ -171,10 +171,10 @@ export class ModuleSetupObsidian extends AbstractObsidianModule implements IObsi
                         await this.core.rebuilder.$fetchLocal(true);
                     } else if (setupType == setupAgain) {
                         const confirm =
-                            "I know this operation will rebuild all my databases with files on this device, and files that are on the remote database and I didn't synchronize to any other devices will be lost and want to proceed indeed.";
+                            "This operation will rebuild all databases with files on this device. Any files on the remote database not synced here will be lost.";
                         if (
                             (await this.core.confirm.askSelectStringDialogue(
-                                "Do you really want to do this?",
+                                "Are you sure you want to do this?",
                                 ["Cancel", confirm],
                                 { defaultAction: "Cancel" }
                             )) != confirm
--- a/src/modules/features/ModuleSetupObsidian.ts
+++ b/src/modules/features/ModuleSetupObsidian.ts
@@ -142,7 +142,7 @@ export class ModuleSetupObsidian extends AbstractObsidianModule implements IObsi
                     const setupJustImport = "Don't sync anything, just apply the settings.";
                     const setupAsNew = "This is a new client - sync everything from the remote server.";
                     const setupAsMerge = "This is an existing client - merge existing files with the server.";
-                    const setupAgain = "This is a new server - reset remote server and use local files.";
+                    const setupAgain = "Initialise new server data - ideal for new or broken servers.";
                     const setupManually = "Continue and configure manually.";
                     newSettingW.syncInternalFiles = false;
                     newSettingW.usePluginSync = false;
--- a/src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
+++ b/src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
@@ -463,7 +463,7 @@ export class ObsidianLiveSyncSettingTab extends PluginSettingTab {
                         // And modified.
                         this.plugin.confirm.askInPopup(
                             `config-reloaded-${k}`,
-                            `The setting "${getConfName(k as AllSettingItemKey)}" being in editing has been changed from somewhere. We can discard modification and reload by clicking {HERE}. Click elsewhere to ignore changes`,
+                            `The setting "${getConfName(k as AllSettingItemKey)}" was modified from another device. Click {HERE} to reload settings. Click elsewhere to ignore changes`,
                             (anchor) => {
                                 anchor.text = "HERE";
                                 anchor.addEventListener("click", () => {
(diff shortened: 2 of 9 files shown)


src/modules/features/ModuleSetupObsidian.ts
```diff
@@ -139,11 +139,11 @@ export class ModuleSetupObsidian extends AbstractObsidianModule implements IObsi
...
                     newSettingW.additionalSuffixOfDatabaseName = `${"appId" in this.app ? this.app.appId : ""}`;
-                    const setupJustImport = "Just import setting";
-                    const setupAsNew = "Set it up as secondary or subsequent device";
-                    const setupAsMerge = "Secondary device but try keeping local changes";
-                    const setupAgain = "Reconfigure and reconstitute the data";
-                    const setupManually = "Leave everything to me";
+                    const setupJustImport = "Don't sync anything, just apply the settings.";
+                    const setupAsNew = "This is a new client - sync everything from the remote server.";
+                    const setupAsMerge = "This is an existing client - merge existing files with the server.";
+                    const setupAgain = "This is a new server - reset remote server and use local files.";
```
vrtmrz, line 145: "This is a new server":  Can you make it a bit more nuanced, like ‘this client creates new server data’?
`This is for a new or broken server`? Is it too long?

src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
```diff
@@ -922,22 +922,25 @@ Store only the settings. **Caution: This may lead to data corruption**; database
...
+                    .setName("Manual setup").addButton((text)
+                    .setDesc("Not recommended, but useful if you don't have a Setup URI") => {
+                        text.setButtonText("Start").onClick(async () => {
+                            await this.enableMinimalSetup();
+                        });
                 });
                 new Setting(paneEl)
-                    .setName("Enable LiveSync on this device as the setup was completed manually")
+                    .setName("Enable LiveSync")
+                    .setDesc("Only enable this after configuring either of the above two options.")
```
vrtmrz, line 943: Please add the case that completed all configuration manually.

src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
```diff
@@ -1526,8 +1533,8 @@ However, your report is needed to stabilise this. I appreciate you for your grea
 
             void addPanel(paneEl, "Fetch settings").then((paneEl) => {
                 new Setting(paneEl)
-                    .setName("Fetch tweaks from the remote")
-                    .setDesc("Fetch other necessary settings from already configured remote.")
+                    .setName("Fetch config from server")
```
vrtmrz, line 1536: Please keep `remote` or change something good: we can serve some database (or bucket) on a single server. (It used to be Database, before implemented bucket sync) However, as you fixed, this is a bit ambiguous.

src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
vrtmrz, line 1: Please keep `remote` or change something good: we can serve some database (or bucket) on a single server. (It used to be Database, before implemented bucket sync) However, as you fixed, this is a bit ambiguous.

src/modules/features/SettingDialogue/settingConstants.ts
```diff
@@ -105,48 +105,50 @@ export const SettingInformation: Partial<Record<keyof AllSettings, Configuration
     },
     syncOnSave: {
         name: "Sync on Save",
-        desc: "When you save a file, sync automatically",
+        desc: "Forces a sync of the file when it's saved.",
```
vrtmrz, line 108: Sorry for the ambiguity, but I'm glad your PR uncovered it. This simply starts synchronisation when we save the file, so the file that is synchronised is not limited to this file.

src/modules/features/SettingDialogue/settingConstants.ts
```diff
@@ -105,48 +105,50 @@ export const SettingInformation: Partial<Record<keyof AllSettings, Configuration
...
     },
     disableMarkdownAutoMerge: {
-        name: "Always resolve conflicts manually",
-        desc: "If this switch is turned on, a merge dialog will be displayed, even if the sensible-merge is possible automatically. (Turn on to previous behavior)",
+        name: "Always prompt merge conflicts",
+        desc: "Should we prompt you for every single merge, even if we can safely merge automatcially?",
     },
     writeDocumentsIfConflicted: {
-        name: "Always reflect synchronized changes even if the note has a conflict",
+        name: "Sync changes even if the file has a conflict.",
```
vrtmrz, line 151: If the file has a conflict, the wording `sync` is not correct. The original `Synchronised` was totally ambiguous. How about `Apply the latest transferred change even if the file has a conflict.` or `Write the latest transferred change even...`?

src/modules/features/SettingDialogue/settingConstants.ts
```diff
@@ -259,7 +261,7 @@ export const SettingInformation: Partial<Record<keyof AllSettings, Configuration
     },
     useCustomRequestHandler: {
         name: "Use Custom HTTP Handler",
-        desc: "If your Object Storage could not configured accepting CORS, enable this.",
+        desc: "Enable this if your Object Storage doesn't accept CORS",
```
vrtmrz, line 264: How about `Enable this if your Object Storage is not configurable about CORS`?

vrtmrz:
Thanks to your rewrite, I am very aware that the original English text was ambiguous and difficult to understand. It was misleading and indeed deceptive. Your rewrite has made it much clearer.
I have added many things in greedily and would be grateful for good ideas for them!

src/modules/features/ModuleSetupObsidian.ts
Volkor3-16: I agree, my wording was a little simple there - I was trying to keep them as short as possible to make it easy to read at a glance, even if it left out some information.
I've changed it to `Initialise new server data - ideal for new or broken servers.`
There's probably a even better way of wording it, but this seems much better for now :)

src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
Volkor3-16: I've clarified the strings to say "remote server" - I feel like the word remote on its own can be a little bit confusing - the word server is more generic.

src/modules/features/SettingDialogue/settingConstants.ts
Volkor3-16: changed to `Starts synchronisation when a file is saved.`

src/modules/features/SettingDialogue/settingConstants.ts
Volkor3-16: Rewrote these again to the following. 
```
name: "Apply Latest Change if Conflicting",
desc: "Enable this option to automatically apply the most recent change to documents even when it conflicts",
```

src/modules/features/SettingDialogue/ObsidianLiveSyncSettingTab.ts
```diff
@@ -922,22 +922,25 @@ Store only the settings. **Caution: This may lead to data corruption**; database
...
                             eventHub.emitEvent(EVENT_REQUEST_OPEN_SETUP_URI);
                         });
                     });
 
-                new Setting(paneEl).setName("Minimal setup").addButton((text) => {
-                    text.setButtonText("Start").onClick(async () => {
-                        await this.enableMinimalSetup();
-                    });
+                new Setting(paneEl)
+                    .setName("Manual setup").addButton((text)
```
vrtmrz, line 935: Please fix here. Probably just only the line has been moved during your work!

```
  new Setting(paneEl)
                    .setName("Connect with Setup URI")
                    .setDesc("This is the recommended method to set up Self-hosted LiveSync with a Setup URI.")
                    .addButton((text) => {
                        text.setButtonText("Use").onClick(() => {
```

vrtmrz:
Thanks for polishing it up! I think that it has got very clear!

Now for other minor things, I hope to ask you for two additional works.
- Fix this [minor issue](https://github.com/vrtmrz/obsidian-livesync/pull/543#discussion_r1867428209)
- run `npx run pretty` to prettify.

I am looking forward to merge your awesome contribution!

Volkor3-16:
Done and done!

One thing to note though, it seems like you set it up to use `crlf` - however the entire project is in `lf` formatting, so I've changed the prettierrc file to use that.

Realistically, it doesn't matter for compatibility, except for ancient versions of windows (winxp or earlier)

vrtmrz:
Thank you for all your fine work! I am so happy to have this merged!
The docs will also be updated soon to reflect the title of the settings!

PR #562, opened 2024-12-28
WIP: Consistent/Polished wording in Docs & Settings

RyderForReal:
First of - thanks a lot for your work on this project. I recently started using it since I generally prefer to self-host things and it has been incredibly useful.

I figured that I'd make the titles and descriptions in the docs and settings more consistent and reword some cases of awkward phrasing. I'm still working on this PR, but was wondering if you had a preference for which kind of spelling to use (American English or British English). I'd just run a quick search & replace when I'm done with my work to make this consistent as well.

Thanks in advance.


RyderForReal:
Notes to self:

- Fix the translation keys
- Parity between settings.md and the actual settings

RyderForReal:
Question regarding the cache size setting (Advanced -> Memory Cache):

The old description is "(Mega chars)" (see https://github.com/vrtmrz/obsidian-livesync/blob/0629bc04bb69412e1cb7b518d6b491425a37b6e9/src/modules/features/SettingDialogue/settingConstants.ts#L90), implying million characters (mega = million).

However, when reading through the code that refers back to this setting, specifically LRUCache, I don't see a mention of million. Am I correct in assuming that the "million" can be removed from the description?

RyderForReal:
Thanks for the document and apologies for the late response on my end as well (I don't check GitHub often since I usually work on my private Git instance).

I'm quite busy for the next few months, so I won't be able to commit a lot (if any) time to this PR, but let me know what further changes are needed on my end (aside from implementing the changes in the doc and resolving merge conflicts) and I'll work on them once I find the time. If it makes reviewing and merging easier for you, I could also split this into a PR for the README, PR for UI i18n, etc.

PR #566, opened 2025-01-06
Add a specialized docker container for obsidian-livesync couchdb

oleduc:
I created a docker image that automatically configures couchDB based on the latest version of couchdb-init.sh

https://github.com/oleduc/docker-obsidian-livesync-couchdb

### DIFF
--- a/docs/setup_own_server.md
+++ b/docs/setup_own_server.md
@@ -105,6 +105,17 @@ export username=johndoe
 export password=abc123
 deno run -A https://raw.githubusercontent.com/vrtmrz/obsidian-livesync/main/utils/flyio/generate_setupuri.ts
 ```
+Or you can use docker
+```bash
+docker run \
+  -e hostname=https://example.com \
+  -e database=obsidiannotes \
+  -e username=johndoe \
+  -e password=abc123 \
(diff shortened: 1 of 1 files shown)


Quadrubo:
I don't think this should remove the instructions on how to do it using the official couchdb docker container. I would still want to use the official one. Maybe add this as an alternative step?

captainhook:
We can use Docker's ENTRYPOINT to configure a default execution for each time the container runs - I don't think there's harm in it running each time it starts. Or maybe someone else can think of a more efficient way.

vrtmrz:
Sorry for being late!
It pains me to say this, given your valuable contribution, but @Quadrubo has already said, we should be very cautious and discreet with adding new dependencies.
I understand that this may be disappointing, especially since the thing is quite useful. However, just pulling an image from the registry is quite fragile against transparency, if it is not an `official` one -- without giving enough eyeballs. We should try to maximise the eyeballs on the code we depend on. The best way is to use the official and a small script to automate the process, and be free to/for all your changes. (@ben-ba mentioned). If we can choose that, we choose so.
I appreciate your understanding on this matter. However, I close the PR without merging.

PR #704, opened 2025-08-29
Add Docker Compose troubleshooting for own server setup

Gron-HD:
Added another option for the init command that passes the variables directly. When i tried the command above i got the mentioned error.


vrtmrz:
Thank you for your PR! This makes the process more robust!
Let me appreciate you for improving the project!

PR #801, opened 2026-02-15
Add step to toggle LiveSync in setup instructions

yominater:
Reminder to toggle LiveSync on

As a not very experienced user, I was confused when client was not syncing. Adding this so people know the option exists


vrtmrz:
Thank you for your PR! And, so sorry for the late response.

However, if I may be so bold, I think this depends on the user’s preference. (For example, a mobile user will want to use periodic sync).
Instead of this, how about changing this instruction to the link to [Sync settings](https://github.com/vrtmrz/obsidian-livesync/blob/main/docs/quick_setup.md#sync-settings)?

yominater:
That would be great! I just wanted something to let them know there's more to setting up syncing

vrtmrz:
Well, I also think so!

Would you mind if I asked you to amend the pull request as such? Then I will merge your contribution!

vrtmrz:

PR #802, opened 2026-02-16
fix: handle "File already exists" for .md files in writeFileAuto

waspeer:
## Problem

During startup sync (UPDATE STORAGE), LiveSync runs up to 10 operations concurrently. In this scenario, `getAbstractFileByPath` can return `null` for `.md` files whose Obsidian vault index entry hasn't been populated yet — even though the file already exists on disk. This causes `vault.create()` to throw `"File already exists."`, which surfaces as repeated errors in the log:

```
Error while UPDATE STORAGE
Error: File already exists.
```

The final summary shows these as failures:
```
UPDATE STORAGE All done: DONE:18, FAILED:9
```

Even though the files are actually fine on disk, the errors are confusing and the failed-count in the summary is misleading.

## Root cause

In `writeFileAuto`, when `getAbstractFileByPath` returns `null`, the code calls `vault.create()` for `.md` files. If a concurrent init operation (or a previous sync pass) already wrote the file to disk before Obsidian's index caught up, `vault.create()` throws.

## Fix

This is the same stale-index issue already identified (and commented on) for non-`.md` files — handled there via `adapterWrite`. This PR extends the same pattern to `.md` files: catch `"File already exists."` and fall back to `adapterWrite`, exactly consistent with the existing approach directly above in the same function.

## Testing

Reproduced by having `.md` files in a vault that LiveSync's internal file sync processes on startup (e.g. files in `.claude/` or similar hidden directories). After this fix, the errors no longer appear.

### DIFF
--- a/src/modules/coreObsidian/ModuleFileAccessObsidian.ts
+++ b/src/modules/coreObsidian/ModuleFileAccessObsidian.ts
@@ -108,7 +108,22 @@ export class ModuleFileAccessObsidian extends AbstractObsidianModule implements 
                 // For safety, check existence
                 return await this.vaultAccess.adapterExists(path);
             } else {
-                return (await this.vaultAccess.vaultCreate(path, data, opt)) instanceof TFile;
+                // The same stale-index issue described above can also happen for .md files during
+                // concurrent initialisation (UPDATE STORAGE runs up to 10 ops in parallel).
+                // getAbstractFileByPath returns null because Obsidian's in-memory index hasn't
+                // caught up yet, but the file already exists on disk — causing vault.create() to
+                // throw "File already exists."
+                // Fall back to adapterWrite (same approach used for non-md files above) so the
+                // file is written correctly without an error.
+                try {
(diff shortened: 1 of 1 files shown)


vrtmrz:
Thank you for your PR! I have not realised this, but this appears to be a good solution!

Unfortunately, as we've made drastic changes in the beta, the merge into main may not reflect these directly.
If it is acceptable for me to port it after the merge, I will proceed with merging it as is and then porting it!

The destination is [here](https://github.com/vrtmrz/livesync-commonlib/blob/d038ee5149ccd853210ad755e6951139105e9de6/src/serviceModules/ServiceFileAccessBase.ts#L105), but I have layered so much abstraction that I am rather reluctant to ask you to read it after such a contribution.

Shall I merge your PR?

vrtmrz:
Having asked, I believe it is time to merge the beta, so I hope to merge this PR into main.
After that, I will port it to beta. Thank you for your contribution!

PR #804, opened 2026-02-21
Add connection warning style logic and settings UI dropdown

A-wry:
Adds a dropdown setting to control how connection error warnings are displayed when the sync server is unreachable (e.g. when disconnected from Tailscale).

Options:
- Banner: shows the full error banner (default, existing behavior)
- Icon only: shows a 🔗❌ icon instead of the full banner  
- Hidden: suppresses connection warnings entirely (not recommended, but has been requested)

This addresses user requests to reduce the permanent error banner that appears during offline use.

Requires: https://github.com/vrtmrz/livesync-commonlib/pull/20


vrtmrz (changes requested):
Thank you for your PR! I think fixing the following things brushes PR up more.
Would you mind if I asked you to consider it, please?

A-wry:
Heyo, I implemented your suggestion for the '\u{200b}Network Error:' prefix and it worked almost too well! It suppressed only the 'Failed to fetch' error, but not subsequent errors that fire as a consequence. I implemented logic for those as well:

- Tagged the `Failed to fetch` error in `ModuleObsidianAPI.ts` with `\u{200b}Network Error:` as the root network error source

- Investigated two downstream errors that fire as a consequence of the network failure:

  - `Failed to initialise the encryption key, preventing replication.` -- after tracing through `ensureReplicatorPBKDF2Salt` -> `ensurePBKDF2Salt` -> `getReplicationPBKDF2Salt`, confirmed this always fails due to a remote connection attempt, so it is safe to tag unconditionally with `\u{200b}`
 
  - `Replication has been cancelled by some module failure` -- rather than tagging this one, `_canReplicate` now checks `_previousErrors` for any `\u{200b}`-prefixed messages before firing, suppressing it when a network error is already the known cause while preserving it for genuine non-network failures
  
- Updated the setting label from "Connection warning style" to "Network warning style" for consistency with the rename

It took a while to track down and make sure the downstream errors were safe to tag--I hope my logic is sufficient!

downstream errors example

vrtmrz:
Thank you! It has turned out splendidly.
It may change slightly when porting to the beta I am about to release, but I will base it on your contribution.
Thank you ever so much for your contribution. I merge this!

PR #817, opened 2026-03-09
fix/env var shell collisions in URI setup

rwbru:
The variables `hostname` and `username` used in the URI setup script potentially collide with shell-reserved variables. On some systems, this makes it impossible to set them via export (e.g. fish shell on arch has `hostname` set to read-only). The other variable names (password, database, etc.) are generic enough to risk collisions with other tools.

This PR renames all environment variables in the URI setup to use a LIVESYNC_ prefix, following the common convention for namespaced environment variables. I rewrote the docs (EN only) accordingly. Fix is tested and works fine. As the variables do not seem to be used as dependencies elsewhere, this should be safe to merge


vrtmrz:

Issue #832, opened 2026-03-22
CLI: sync command fails silently — openReplication returns false


rewse:
## Description

The CLI `sync` command connects to CouchDB successfully (all HTTP requests return 200) but `openReplication` returns `false`, causing the command to fail with no actionable error message.

## Steps to reproduce

```bash
git clone --recursive https://github.com/vrtmrz/obsidian-livesync.git
cd obsidian-livesync
npm install
cd src/apps/cli
npm run build

# Create settings and configure CouchDB connection
npm run --silent cli -- init-settings ~/livesync/settings.json
# (edit settings.json to add couchDB_URI, couchDB_USER, couchDB_PASSWORD, couchDB_DBNAME, passphrase, isConfigured: true)

npm run --silent cli -- ~/livesync --settings ~/livesync/settings.json --verbose sync
```

## Output

```
[Command] sync
(debug) [RemoteService]  [REQ] (GET) https://example.com/db/_local/obsidian_livesync_sync_parameters -> 200
(debug) [RemoteService]  [REQ] (GET) https://example.com/db/ -> 200
(debug) [RemoteService]  [REQ] (GET) https://example.com/db/ -> 200
(debug) [RemoteService]  [REQ] (GET) https://example.com/db/obsydian_livesync_version -> 200
(debug) [RemoteService]  [REQ] (GET) https://example.com/db/_local/obsydian_livesync_milestone -> 200
[Error] Command 'sync' failed
```

## Investigation

I added debug logging and found:

- `appLifecycle.isReady()` → `true`
- `replicator.getActiveReplicator()` → exists (not undefined)
- `activeReplicator.openReplication(settings, false, true, false)` → `false`

CouchDB authentication and connectivity are fine (all 200s), but the replication itself never starts. No `_changes` feed is read and no documents are pulled.

## Additional notes

- The `setup` command also has an issue: it requires full core initialization (`onLoad`) before it can run, but `onLoad` fails on a fresh install because the database isn't configured yet. The README's bootstrap example (`init-settings` → `setup` → `sync`) doesn't work as described.
- `remoteType` is empty after `setup` applies the URI. Setting it manually to `"couchdb"` didn't help.

## Environment

- macOS 26.4 (Apple Silicon / arm64)
- Node.js v25.6.1
- Commit: `2ff60dd`

PR #833, opened 2026-03-22
fix(cli): show actionable error when sync fails due to locked remote DB

rewse:
# PR: Show actionable error when CLI sync fails due to locked remote DB

## Problem

When the remote database milestone has `locked: true` and the CLI device's node ID is not in `accepted_nodes`, the `sync` command fails silently with only a generic library-level message:

```
The remote database has been rebuilt or corrupted since we have synchronized last time.
Fetch rebuilt DB, explicit unlocking or chunk clean-up is required.
Could not connect to server.
```

This message references "Fetch rebuilt DB" which is an Obsidian plugin UI action — the CLI has no equivalent command. The "Could not connect to server" is also misleading since the connection itself succeeds.

## Fix

After `replicate()` returns `false`, check the replicator's `remoteLockedAndDeviceNotAccepted` flag and print a CLI-specific error message directing the user to unlock the database from the Obsidian plugin.

## Testing

Tested with a CouchDB remote that had `locked: true` in its milestone document. Before this fix, the CLI showed only the generic messages above. After this fix, it additionally shows:

```
[Error] The remote database is locked and this device is not yet accepted.
[Error] Please unlock the database from the Obsidian plugin and retry.
```

Ref: #832

### DIFF
--- a/src/apps/cli/commands/runCommand.ts
+++ b/src/apps/cli/commands/runCommand.ts
@@ -22,6 +22,9 @@ export async function runCommand(options: CLIOptions, context: CLICommandContext
         console.log("[Command] sync");
         const result = await core.services.replication.replicate(true);
         if (!result) {
+            // TODO: Standardise the logic for identifying the cause of replication
+            // failure so that every reason (locked DB, version mismatch, network
+            // error, etc.) is surfaced with a CLI-specific actionable message.
             const replicator = core.services.replicator.getActiveReplicator();
             if (replicator?.remoteLockedAndDeviceNotAccepted) {
                 console.error(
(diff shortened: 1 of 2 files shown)


vrtmrz:
Thank you for opening the PR! 

There are several reasons why replication might be blocked, so it might be a good idea to standardise the logic for identifying the cause.

That said, I think this fix is an excellent first step towards addressing that issue. It would be helpful if you could add it to the inline //TODO comment (I occasionally grep it). 

Also, it would be helpful if you could create a reproduction for the CLI tests and write a test to verify that the fix has worked. A shell script would be fine for this.

Sorry for asking you to do some extra work; thank you so much for your contribution!

PR #843, opened 2026-03-31
cli: implement continuous sync daemon mode

andrewleech:
## Summary

The CLI's `daemon` command (the default when no command is given) currently initialises LiveSync, suspends all sync, then blocks indefinitely doing nothing. This PR makes it a functional continuous sync daemon.

- **CouchDB → local filesystem**: uses the `_changes` feed (LiveSync mode) by default, or polling via `--interval <seconds>`
- **Local filesystem → CouchDB**: chokidar file watcher, auto-started via the existing service lifecycle
- **Startup sequence**: mirror scan to reconcile offline changes → restore sync settings → `applySettings()` to start the `_changes` feed
- **Conflict resolution**: remote-wins, logged to stderr
- **Ignore rules**: `.livesync/ignore` file with glob patterns; supports `import: .gitignore` directive to merge gitignore rules; applied to both sync directions
- **Deployment**: systemd user service unit + `install.sh` script in `src/apps/cli/deploy/`

## Usage

```bash
# Continuous sync (LiveSync _changes feed)
livesync-cli /path/to/vault

# Polling every 30 seconds
livesync-cli /path/to/vault --interval 30
```

## `.livesync/ignore` example

```
# ignore editor swap files
*.tmp
.#*

# import gitignore rules (excludes .git/, build/, etc.)
import: .gitignore
```

## Systemd install

```bash
bash src/apps/cli/deploy/install.sh --vault /path/to/vault
```

## Notes

- `processSynchroniseResult` no-op interception is now conditional on non-daemon commands, allowing the default handler to write incoming CouchDB changes to the filesystem in daemon mode
- `suspendAllSync()` clobbers 7 sync settings; these are all captured before the call and restored after the mirror scan via `applySettings()` which runs the full `onSuspending → onResuming → onResumed` lifecycle
- chokidar ^4.0.0 added as a dependency (left external in the Vite build, included in `runtime-package.json` for Docker)
- Ignore rules are loaded before `LiveSyncBaseCore` construction so they are available when `beginWatch()` fires during `onFirstInitialise`

## Test plan

- [ ] `livesync-cli /vault` starts, runs mirror scan, then enters LiveSync mode
- [ ] Remote change in CouchDB appears on local filesystem
- [ ] Local file create/modify/delete reflects in CouchDB
- [ ] `--interval 30` polls every 30s instead of using the changes feed
- [ ] `.livesync/ignore` with `*.tmp` prevents `.tmp` files from syncing
- [ ] `import: .gitignore` merges gitignore patterns
- [ ] SIGTERM causes clean shutdown (exit 0)
- [ ] Mirror scan failure at startup exits non-zero
- [ ] `bash src/apps/cli/test/test-daemon-linux.sh` passes (requires CouchDB env vars)


andrewleech:
I've been using this regularly for a couple of weeks now and it seems to be working well.

Joysimple:
I've also used this for this long with my fix on top of your PR and it's indeed a really nice approach.

vrtmrz (approved):
Thank you for the PR! I think that this is a brilliant improvement!
I have implemented another feature: `mirror`, so I will resolve the conflicts.

vrtmrz:
I have merged the changes regarding the separation of `databasePath` and `vaultPath`. 
Guessing that this should resolve the issue where unintended files were being collected in the events... (Of course, `.ignore` will also be useful).

Since I do not have a bare-metal Linux setup, would you mind if I asked you to check the behaviour in your environment, please?

Joysimple:
The latest version of merged main into this PR doesn't work due to a regression @vrtmrz 

Here is a proposed fix (without it, fails with vaultPath being undefined)

```
diff --git a/src/apps/cli/main.ts b/src/apps/cli/main.ts
index 065fae0..41bccf0 100644
--- a/src/apps/cli/main.ts
+++ b/src/apps/cli/main.ts
@@ -265,6 +265,10 @@ export async function main() {
         process.exit(1);
     }

+    // Resolve vault path (defaults to database path unless specified for mirror command)
+    const vaultPath =
+        options.command === "mirror" && options.commandArgs[0] ? path.resolve(options.commandArgs[0]) : databasePath;
+
     // Resolve settings path
     const settingsPath = options.settingsPath
         ? path.resolve(options.settingsPath)
@@ -273,6 +277,7 @@ export async function main() {

     infoLog(`Self-hosted LiveSync CLI`);
     infoLog(`Database Path: ${databasePath}`);
+    infoLog(`Vault Path: ${vaultPath}`);
     infoLog(`Settings: ${settingsPath}`);
     infoLog("");

@@ -364,11 +369,7 @@ export async function main() {
     const core = new LiveSyncBaseCore(
         serviceHubInstance,
         (core: LiveSyncBaseCore<NodeServiceContext, any>, serviceHub: InjectableServiceHub<NodeServiceContext>) => {
-            const mirrorVaultPath =
-                options.command === "mirror" && options.commandArgs[0]
-                    ? path.resolve(options.commandArgs[0])
-                    : databasePath;
-            return initialiseServiceModulesCLI(mirrorVaultPath, core, serviceHub, ignoreRules, watchEnabled);
+            return initialiseServiceModulesCLI(vaultPath, core, serviceHub, ignoreRules, watchEnabled);
         },
         (core) => [
             // No modules need to be registered for P2P replication in CLI. Directly using Replicators in p2p.ts
```

andrewleech:
Thanks @vrtmrz for taking the first pass at the rebase and for the `mirror` feature — that separation of `databasePath` and `vaultPath` made the ignore rules much cleaner to wire up. And thanks @Joysimple for catching the `vaultPath undefined` regression after that rebase, and for the earlier find on the `stat.size` mismatch — both are fixed here.

Tested on bare-metal Linux (Manjaro, systemd user service). Full startup ran cleanly:
- CouchDB → PouchDB replication completed
- Mirror scan: 4508 files, DONE, FAILED:0
- `[Daemon] LiveSync active`

Two warnings on startup, both pre-existing upstream:
- `Error in Setting Save Event` — the new `remoteConfig` migration runs every startup because the daemon restores the pre-startup settings file on exit (intentional: sync settings must survive a crash). Harmless.
- `TimeoutNegativeWarning: -181` — from PouchDB replication internals, nothing to do with daemon code.

Branch is rebased onto current `origin/main` and squashed to 5 commits. The `stat.size` fix is the last one, with `Co-authored-by: @Joysimple`. All 70 unit tests pass.

vrtmrz (approved):
I really thank you for your merging! It seems perfect!
And, sorry for octagonal-wheels, I think a more graceful feature detection is required at that. 
Anyway, your contribution is perfectly fine!

PR #871, opened 2026-05-02
feat: Add diff navigation buttons for Document History

### LINKED ISSUE #870: Add navigation buttons for jumping to diffs in Document History
### Abstract
For long documents, it can be very tedious to manually scroll through the entire text to locate the highlighted `history-added` or `history-deleted` blocks when viewing the Document History. This is a **Feature Request** (with an incoming Pull Request) to add navigation buttons to solve this.

### Expected behaviour
When "Highlight diff" is enabled in the Document History modal, there should be an easy way to jump between the differences without manually scrolling. I expect to see "Previous" and "Next" buttons that automatically scroll the view to the next modified text block, along with an indicator showing the current diff index (e.g., "1/5").

### Actually happened
Currently, users must manually scroll through the entire document to hunt for the green/red highlighted blocks, which is highly inefficient for long notes.

### Reproducing procedure
1. Open a long note that has been synced and has multiple revisions.
2. Select "Show history" from the command palette.
3. Check "Highlight diff".
4. Notice that you have to manually scroll to find the highlighted changes.

### LiveSync version
Latest

SeleiXi:
close #870 

## Summary

Adds navigation button for the Document History that allow users to quickly jump to diff content.

## Motivation

When viewing long documents with the diff highlighting feature, it can be tedious
to manually scroll through the entire content to find changes between revisions.
This enhancement adds simple up/down navigation buttons to jump directly to the
previous or next diff block.

## Changes

- Added "Prev" and "Next" navigation buttons to the Document History modal
- Added a diff position indicator showing current position
- Buttons are only visible when "Highlight diff" is enabled

## Demo
Click "Next" and jump to the corresponding diff
image

### DIFF
--- a/src/modules/features/DocumentHistory/DocumentHistoryModal.ts
+++ b/src/modules/features/DocumentHistory/DocumentHistoryModal.ts
@@ -223,6 +223,9 @@ export class DocumentHistoryModal extends Modal {
         }
         // Reset diff navigation after content changes
         this.resetDiffNavigation();
+        if (this.showDiff) {
+            this.navigateDiff("next");
+        }
     }
 
     /**
--- a/src/modules/features/DocumentHistory/DocumentHistoryModal.ts
+++ b/src/modules/features/DocumentHistory/DocumentHistoryModal.ts
@@ -66,6 +66,11 @@ export class DocumentHistoryModal extends Modal {
     currentDeleted = false;
     initialRev?: string;
 
+    // Diff navigation state
+    currentDiffIndex = -1;
+    diffNavContainer!: HTMLDivElement;
+    diffNavIndicator!: HTMLSpanElement;
+
     constructor(
         app: App,
         core: LiveSyncBaseCore,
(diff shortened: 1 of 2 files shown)


SeleiXi:
There's some further commits I wanna implement

1. Merge modifications within one minute into a single version
  Current Status Analysis: PouchDB (the underlying database) is designed such that every put operation generates a new _rev.
  
For this one, may I know which implementation is the way you want:
- UI-Level Filtering: When displaying history records in DocumentHistoryModal or GlobalHistory, if the time difference between two adjacent versions is less than 60 seconds, display only one version (or collapse them) on the interface.
  
- Database-Level Merging (More difficult): When saving files (EntryManagerImpls.ts), check the mtime of the previous version. If it is less than 60 seconds, remove the old version before performing a put for the new version. However, this approach is risky as it might disrupt synchronization logic or lead to version conflicts.
 
---

2.  implementing "Auto-Locate" to automatically scroll to the first modification as soon as a revision is loaded.
- After showExactRev finishes loading the content, automatically trigger MapsDiff("next") once. This ensures that as soon as a user opens a specific version, their viewport immediately focuses on the modified section.


btw, I would like to open a new PR to support History Search Functionality (Recent 100 Versions / Diff Retrieval). 

As well as a new PR for DocumentHistoryModal Window Maximize.
- Add a fullscreen/maximize button to the header of the Modal.

SeleiXi:
> implementing "Auto-Locate" to automatically scroll to the first modification as soon as a revision is loaded.

I have already implemented this feature, but I am considering the following two points:

- Toggle Button: Do we need to add a toggle button/setting that allows users to enable or disable "Auto-Locate"?
- Scroll Animation: Currently, the automatic positioning has a smooth scrolling effect. For example, when switching between two versions, there might be a visible scrolling transition from line 1 all the way down to line 1000. Should we remove this animation and jump directly to the target line instead?

vrtmrz:
Thank you for opening the PR!
I think this feature may be quite useful!

I was thinking of asking you for a small favour, but I realised your implementation seems to be working fine.
And, indeed, I have tested it just now, and it seems to be working so well. Hence, I will merge it for now! Thank you for your contribution!

Regarding the question about the feature extension, it might be better to implement this as a new feature. The reason is that I suspect extending Modal would make things rather complicated when combined with DOM manipulation. It might be better to implement it as a new pane with Svelte.

That said, I do think this PR is a fully-fledged feature addition in its own right, and I also think your new proposal is a good one. Why don't we do them step by step?

PR #876, opened 2026-05-08
fix: Suppress network error notice popups when "NetworkWarningStyle" set to "Hidden"

### LINKED ISSUE #873: [BUG] Set NetworkWarningStyle to "Hidden" does not suppress "Network Error: Failed to fetch" Notice popups
### Abstract
The `networkWarningStyle` setting (Hidden/Icon) does not suppress "Network Error: Failed to fetch" Notice popups. These banners continue to appear even when the user has configured them to be hidden or shown only as icons.

Root Cause
1. ModuleLog.ts's __addLog method triggers new Notice() without checking networkWarningStyle.
2. Many network-related errors in RemoteService.ts lack the MARK_LOG_NETWORK_ERROR tag.

The networkWarningStyle setting introduced in PR #804 (addressing #770) is incomplete. While it successfully hides the persistent editor banner, it fails to suppress the transient Notice popups.

### Expected behaviour

When `General -> Network Warning Style` is set to `Hide completely` or `Show icon only`, transient network errors (like fetch failures) should not trigger a system-level `Notice` (the popup in the top-right corner).

### Actually happened

A `Notice` containing "Network Error: Failed to fetch: ..." pops up frequently, especially on mobile devices with unstable connections, even if the warning style is set to "Hidden".

### Reproducing procedure

1. Go to **Settings -> LiveSync -> General**.
2. Set **Network Warning Style** to **Hide completely**.
3. Cause a network interruption (e.g., disable Wi-Fi/Data or use an unstable proxy).
4. Observe that the "Network Error: Failed to fetch" popup still appears.

### LiveSync version
0.25.57

### Plug-in log
```
Network Error: Failed to fetch: Error:Requiest failed. The network connection was lost.
```

### Screenshot

Image
Image

Image

SeleiXi:
Fixes #873.

This updates network-error handling so errors tagged with `MARK_LOG_NETWORK_ERROR` respect the configured network warning style more consistently.

Changes:
- Tag the direct offline replication error (`Network is offline`) as a network error.
- Suppress Obsidian `Notice` popups for tagged network errors unless `networkWarningStyle` is set to the full banner mode.
- When `networkWarningStyle` is `hidden`, also avoid showing the tagged network error in the editor status text and temporary editor log history.
- Keep the log entry itself available in the normal log list/log file path for manual inspection.

This PR keeps network errors logged, but prevents them from surfacing as popups when the user selected a less intrusive network warning style. 

It is noticed that errors like [Network Error: Failed to fetch...](https://github.com/vrtmrz/livesync-commonlib/blob/97530553a63dc206ea3fb7ef60721cabda6c74cc/src/services/base/RemoteService.ts#L252C20-L252C20) is already labelled with a prefix {MARK_LOG_NETWORK_ERROR}. Thus, this PR ables to prevent them from alerting.


vrtmrz:
Thank you for the PR! 
To be honest, the problem is that I was actually telling a massive lie on line L252 of `src/lib/src/services/base/RemoteService.ts`. (it is [here](https://github.com/vrtmrz/livesync-commonlib/blob/3e8520dd4702432c9e28c29629044937a1e40156/src/services/base/RemoteService.ts#L252), comment says `Do not show notice`, but without the second argument, it would be handled as `LOG_LEVEL_NOTICE`  )
Passing `LOG_LEVEL_INFO` to the `this.showError` may fix the situation.

It would be a shame to let your kind offer to contribute go to waste, so if you do not mind, could you please fix this and submit a pull request?

SeleiXi:
I've submitted the PR [here](https://github.com/vrtmrz/livesync-commonlib/pull/33). Thanks for your reviewing!

SeleiXi:
I've updated the submodule pointer to the latest version of [livesync-commonlib](https://github.com/vrtmrz/livesync-commonlib). Thank you so much for fixing this issue!

PR #858, opened 2026-04-22
Improve i18n coverage and runtime language updates

52sanmao:
## Summary
- localize additional settings panes, setup dialogs, maintenance flows, and user-facing notices across supported locales
- make the plugin follow the active Obsidian language by default and apply display language changes immediately in the settings UI
- regenerate baked locale assets from YAML sources, add an i18n coverage check command, and update the commonlib submodule pointer

## Test plan
- [x] npm run bakei18n
- [x] npm run tsc-check
- [x] npm run build
- [x] Deploy the built plugin locally and verify translated UI updates in Obsidian

## Dependency
- Depends on vrtmrz/livesync-commonlib#28

PR #878, opened 2026-05-09
Improve i18n coverage and runtime language updates

52sanmao:
﻿## Background

This reworks the i18n/runtime-language update based on the maintainer feedback in #858.

The goal is to localise the setup/settings UI without scattering translation logic across every call site, and without losing the stable typed values returned from dialog flows.

## Purpose

- Keep setup and settings code focused on stable message keys and typed dialog results.
- Move reusable label translation to shared UI controls and confirm/dialog entry points where possible.
- Make runtime language changes refresh the visible settings UI consistently.
- Avoid translating language names in the language selector, so each language is shown in its native form.

## Changed Files

- `src/lib`
  - Updates the submodule to the companion shared UI i18n PR.
  - This is needed because setup UI components such as `Decision`, `DialogHeader`, `InfoNote`, `InputRow`, `Option`, `Check`, `Guidance`, and `Password` should translate known keys while keeping literal/dynamic strings safe.

- `src/modules/features/SetupWizard/**`
  - Replaces remaining user-facing English labels/messages with i18n keys.
  - Adds explicit result types for wizard dialogs so `openWithExplicitCancel` keeps the expected result shape instead of relying on broad casts.
  - Keeps cancellation and connection-test messages localisable.

- `src/modules/services/ObsidianConfirm.ts` and `src/modules/coreObsidian/UILib/dialogs.ts`
  - Centralises translation for common confirm/dialog labels.
  - Maps translated button labels back to the original values before returning, so callers still receive stable typed values.

- `src/modules/features/SettingDialogue/**`
  - Localises settings pane titles and option labels shown through the settings UI.
  - Updates the display-language save path so changing the language refreshes the settings UI.
  - Keeps language names as native display names instead of translating them through i18n keys.

- `src/modules/features/SetupManager.ts` and related tests
  - Uses the explicit wizard result types when opening setup dialogs.
  - Keeps setup flow control typed and avoids relying on generic result casts.

- `src/serviceFeatures/**` and red-flag related modules/tests
  - Localises user-facing warning/result labels that are surfaced through shared dialog/confirm flows.
  - Keeps tests aligned with the typed return values used after label translation.

## Verification

- `npm run tsc-check`
- `npm run svelte-check`
- `npm run build`
- `git diff --check`
- `git -C src/lib diff --check`

## Compatibility / Notes

- Dialog callers should continue to receive the original stable return values, not translated display strings.
- The language selector intentionally shows native language names instead of translating language names.
- This PR depends on the companion commonlib PR: https://github.com/vrtmrz/livesync-commonlib/pull/32


vrtmrz:
Thank you very much for this contribution. 
I am so sorry to inform this, but the translation boundary changed while v1.0.0 was being prepared: LiveSync now owns its application catalogue and dialogues, and consumes Commonlib as a published package. 

A mechanical rebase would preserve the former dependency in the wrong place. 
Would you be willing to update this PR against the current `main`, keeping the LiveSync-owned parts here and removing the former Commonlib catalogue or gitlink dependency, please? 
I am sorry that this requires extra work after your contribution.

Issue #888, opened 2026-05-11
[BUG]Sync sometimes doesn't resume after app returns from background on iOS - requires force quit and reopen


SeleiXi:
On iOS, sometimes the plugin doesn't automatically start syncing when returning the Obsidian app from background. The user must fully quit the app and reopen it for sync to begin.

After investigating the code, I found the possible issue is in `ModuleObsidianEvents.ts` in the `watchWindowVisibilityAsync()`:

```  
if (isHidden) {
      await this.services.appLifecycle.onSuspending();
  } else {
      if (this.services.appLifecycle.isSuspended()) return;
      if (!this.hasFocus) return;  // <-- Problem: this condition blocks resume
      await this.services.appLifecycle.onResuming();
      await this.services.appLifecycle.onResumed();
  }
```
  The hasFocus flag is managed by focus/blur DOM events:
```
  this.plugin.registerDomEvent(window, "focus", () => this.setHasFocus(true));
  this.plugin.registerDomEvent(window, "blur", () => this.setHasFocus(false));

  hasFocus = true;  // Initial value assumes focus
```
  Why this fails on iOS:

  1. iOS WebView's focus/blur event handling differs from desktop browsers
  2. When iOS app switches to background, blur event may not fire reliably
  3. When returning from background, focus event may be delayed or not fire at all
  4. The visibilitychange event fires before focus, and at that moment hasFocus may still be false
  5. The if (!this.hasFocus) return; check then blocks onResuming/onResumed from being called
  6. Consequently, periodic sync and LiveSync don't resume

  Impact:

  - onResumed event never triggers
  - ModulePeriodicProcess doesn't resume periodic sync
  - ModuleReplicatorCouchDB._everyAfterResumeProcess() doesn't restart LiveSync
  - User must force quit and reopen the app as a workaround

vrtmrz:
Thank you for opening the issue!
I had not realised this issue, but your investigation seems very reasonable.
This will be fixed in the next version!

vrtmrz:
I have missed the reply! This has been fixed at v0.25.65.
I think it is sorted now, but would you please close this for me if you do not mind?

SeleiXi:
> I have missed the reply! This has been fixed at v0.25.65. I think it is sorted now, but would you please close this for me if you do not mind?

Thank you so much!

PR #914, opened 2026-05-18
fix: Fix an issue about resuming from background on iOS (#888).

vrtmrz:
This PR includes the following things:

### Improved
- Improved the error verbosity on concurrent processing during the start-up process.

### Fixed

- Fix an issue about resuming from background on iOS (#888).


vrtmrz (commented):
I did performed a proper review on today.

PR #921, opened 2026-05-22
Refactor cli bash e2e to deno 

vrtmrz:


vrtmrz:
This will be merged as part of #948

PR #928, opened 2026-05-28
feat(cli): add --vault option for daemon and mirror commands

starskyzheng:
Allow specifying a separate vault directory for .md files, decoupled from the database directory (where PouchDB data lives).

- Add --vault/-V flag to CLI options parser
- Use vaultPath (or fallback to databasePath) for file system operations
- Works with both daemon and mirror commands
- Log vault path alongside database path at startup

### DIFF
--- a/src/apps/cli/main.ts
+++ b/src/apps/cli/main.ts
@@ -61,7 +61,7 @@ Commands:
     info <path>             Show detailed metadata for a file (ID, revision, conflicts, chunks)
     rm <path>               Mark a file as deleted in local database
     resolve <path> <rev>    Resolve conflicts by keeping <rev> and deleting others
-    mirror [vault-path]     Mirror database contents to the local file system (vault-path defaults to database-path)
+    mirror [vault-path]     Mirror database contents to the local file system (vault-path takes precedence over --vault; defaults to vault from --vault / database-path)
     remote-add <name> <connstr>
                             Add a remote configuration from a connection string
     remote-rm <remote-id>    Remove a remote configuration by ID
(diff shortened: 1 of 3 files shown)


vrtmrz:
Thank you for your PR!
I think vault-path and database-path are quite confusing, too. It is mostly fine, but I would just like to ask you for a small favour, please?

src/apps/cli/main.ts
```diff
@@ -290,16 +304,17 @@ export async function main() {
         : path.join(databasePath, SETTINGS_FILE);
     configureNodeLocalStorage(path.join(databasePath, ".livesync", "runtime", "local-storage.json"));
 
+    // Resolve vault path: --vault flag takes priority, otherwise fall back to databasePath
+    // For daemon mode, enable chokidar file watching so the _changes feed picks up events.
+    // mirror runs a single full scan and doesn't need continuous watching.
+    const watchEnabled = options.command === "daemon";
+    const vaultPath = options.vaultPath ? path.resolve(options.vaultPath) : databasePath!;
```
vrtmrz: I think this should respect the arguments of the `mirror` command as something like follows.

```
const vaultPath =
        options.command === "mirror" && options.commandArgs[0]
            ? path.resolve(options.commandArgs[0])
            : options.vaultPath
              ? path.resolve(options.vaultPath)
              : databasePath!;
```

vrtmrz (changes requested):
Shall I ask you to make amendments?
And, please update the README (I always forget this).

starskyzheng:
Thank you for the review! I've updated the code per your suggestion:

Restored the mirror positional argument handling with the priority order you specified: mirror positional arg → --vault flag → databasePath
Updated src/apps/cli/README.md with documentation for the --vault option (in Arguments, Options, and mirror command description)
Let me know if there are any other amendments needed!

vrtmrz:
Thank you for your additional effort!
I am delighted to have been able to merge your contribution!

Issue #939, opened 2026-06-04
Keep replication running while the window is hidden (opt-in, desktop)


migsferro:
## Problem

On desktop, replication is suspended as soon as the Obsidian window becomes hidden (`document.hidden`, i.e. minimised) and only resumes when the window is shown again. Edits made elsewhere do not arrive until you click back into Obsidian, and local edits can sit unsynced while minimised.

Repro (LiveSync mode, two clients on one CouchDB):

1. Minimise Obsidian on client A.
2. Edit a shared note on client B.
3. Client A receives nothing until its window is focused. With Verbose Log on, replication closes right after `document.hidden` turns true.

Source: `ModuleObsidianEvents.watchWindowVisibilityAsync()` calls `appLifecycle.onSuspending()` on hide.

## Affected modes

- LiveSync: the continuous connection is closed (`suspendReplication` -> `closeReplication`).
- Periodic: the periodic timer is stopped (`disablePeriodic`).
- On-Events: no background channel, so unaffected.

## Proposal

A new opt-in setting, "Keep replication active in the background" (`keepReplicationActiveInBackground`), default off, desktop only. When on, window visibility stops driving the suspend, so replication keeps running while the window is minimised. Becoming visible still forces a clean reconnect.

Default behaviour is unchanged. The key is excluded from setup-URI / QR transfer so it never reaches mobile.

## Implementation note

A kept-alive continuous channel can go half-open (PouchDB stays in `paused`/retry while the replicator keeps its `AbortController` set), so a plain reopen on focus would be a no-op. Becoming visible therefore forces a teardown before reopening, which restores the default's reset-on-visibility. Recovery during a long hidden period (no visibility change to trigger it) would need a liveness watchdog, left out to keep this change small. Net effect: best-effort background sync that recovers on focus and is never worse than the default.

## Status

A PR is ready across both repos if you are open to it:

- `livesync-commonlib`: the setting definition (type, default, label, QR index `-1`).
- `obsidian-livesync`: the guard in the visibility handler, a desktop-only toggle in the Sync pane shown for LiveSync and Periodic, a `docs/settings.md` entry, and unit tests for the handler. Stalled-channel recovery verified manually on Windows (block the connection past the heartbeat while hidden, recovers within ~1s of focus).

Passes the repo's prettier, eslint and tsc on the changed lines.

## Questions

1. Are you open to this as an opt-in, default-off, desktop-only setting?
[...]

Issue #51, opened 2022-03-13
In insider 1.1.1, the option icon is only gears.


rakugakijin:

vrtmrz:
Thank you for making the issue.
May I get a screenshot?

rakugakijin:
Thank you. It was revived.![image](https://user-images.githubusercontent.com/65814903/158273709-60da033a-b9cf-4a1d-9503-ec1226b834bc.jpeg)

PR #949, opened 2026-06-08
feat: opt-in desktop setting to keep replication active in the background

migsferro:
Implements #939. The commonlib setting definition merged in vrtmrz/livesync-commonlib#51; `src/lib` is bumped to commonlib main (`e98d929`, the #51 merge) so this branch builds against current upstream.

## Summary

On desktop, replication is suspended as soon as the Obsidian window becomes hidden (`document.hidden === true`, i.e. minimised) and only resumes when the window is shown again. For the modes that maintain a background channel this stalls background sync: edits made elsewhere do not arrive until the user focuses Obsidian, and local edits can sit unsynced while minimised.

This adds an opt-in, desktop-only setting, **Keep replication active in the background** (`keepReplicationActiveInBackground`, default `false`). When enabled on desktop, window visibility stops driving the replication suspend/resume lifecycle, so replication keeps running while the window is minimised or hidden. Default behaviour is unchanged, and the setting is ignored on mobile (where suspend-on-hidden preserves battery).

## Which modes this affects

`watchWindowVisibilityAsync()` calls `appLifecycle.onSuspending()` on hide. That suspends:

- LiveSync (continuous): closes the live replication connection.
- Periodic: stops the periodic timer.
- On-Events: no background channel, so nothing meaningful is suspended.

So the stall affects LiveSync and Periodic. The setting is general, and the UI shows it only for those two modes.

## Approach

Opt-in guard in `watchWindowVisibilityAsync()`. Pending file events are still committed (file handling unchanged). On hide the suspend is skipped so the channel keeps running. On becoming visible a teardown then reopen is forced, so a stalled background channel is always replaced:

```ts
await this.services.fileProcessing.commitPendingFileEvents();

const keepActiveInBackground =
    this.settings.keepReplicationActiveInBackground &&
    (this.settings.liveSync || this.settings.periodicReplication) &&
    !this.services.API.isMobile();

if (isHidden) {
    if (!keepActiveInBackground) await this.services.appLifecycle.onSuspending();
} else {
    if (this.services.appLifecycle.isSuspended()) return;
    // Only the continuous (LiveSync) channel can stall but stay non-terminal: PouchDB emits
    // paused/retry while the replicator keeps `this.controller` set, so a plain reopen no-ops
    // (openContinuousReplication early-returns). Force a teardown first so becoming visible always
    // re-establishes a fresh channel. Gated on liveSync because Periodic mode has no such channel.
    if (keepActiveInBackground && this.settings.liveSync) {
        await this.services.appLifecycle.onSuspending();
    }
    await this.services.appLifecycle.onResuming();
    await this.services.appLifecycle.onResumed();
}
```

Design notes:

- The guard is in the visibility handler only. `onSuspending()` is also called by `ControlService.applySettings()` and real lifecycle suspends; guarding here scopes the change to the window-hidden trigger and leaves manual suspend, settings-apply, and shutdown/unload untouched.
- Why force a teardown on becoming visible. Continuous replication runs PouchDB `{ live: true, retry: true, heartbeat: 30000 }`, so transient drops self-heal. A non-terminal stall (PouchDB paused/retrying, or a half-open socket after sleep or NAT rebind) keeps the `for await` loop alive, so `this.controller` stays set and `openContinuousReplication` early-returns. A plain reopen would no-op on exactly the channel that needs replacing. Default behaviour resets this implicitly by closing on every hide; since this feature skips that, the resume forces the teardown so visibility recovery is never worse than default. The teardown is gated on `liveSync` only, since the periodic timer has no half-open state and just resumes normally.
- Why the same-tick teardown then reopen does not double-open. `closeReplication()` aborts and nulls the controller synchronously (`controller.abort(); this.controller = undefined`), and the abort is observed by the replicator's `for await` loop even when stalled (the signal is raced into the inbox `pick()`, so a half-open iterator is woken and its `finally` calls `sync.cancel()`). The reopen is deferred: `_everyAfterResumeProcess` wraps it in `fireAndForget(async () => { await isReplicationReady(...); openReplication(...) })`, and `openReplication` awaits `initializeDatabaseForReplication()` before creating a new sync. So the synchronous teardown always completes before a new channel exists. This is the same suspend, guard, resume sequence `ControlService.applySettings()` already runs.
- Latent sharp edge, noted but not changed here. `processSync`'s `finally` clears the shared `this.controller` rather than the local one it created, so in principle an old sync's teardown finishing after a reopen has set a new controller could abort the fresh channel. In this flow the two awaited gates above make the old (microtask-only) teardown win the race in practice, so the PR does not introduce a regression. It is a pre-existing robustness wrinkle in `processSync` worth a separate optional hardening (have the `finally` clear only the controller it owns). The manual test below asserts exactly one live sync after rapid restore cycles.
- Known limitation, documented. While the window stays hidden there is no visible transition to trigger recovery, so a stall during a long hidden period is repaired on the next focus, which is parity with today. Covering that fully would need a liveness watchdog, left as a deliberate follow-up to keep this change small. Most transient drops recover on their own via `retry: true`; only a rare non-terminal stall needs the focus boundary.

## Changes

- `ModuleObsidianEvents.ts`: the opt-in guard above. Platform is read through `services.API.isMobile()` (the plugin's own service layer), not `Platform` directly.
- `PaneSyncSettings.ts`: desktop-only toggle in the Sync settings pane, shown only for LiveSync and Periodic. Same `services.API.isMobile()` check, matching the existing call in `ObsidianLiveSyncSettingTab`.
- `docs/settings.md`: document the setting under Sync Settings.
- `ModuleObsidianEvents.unit.spec.ts`: 8 tests covering the suspend-skip (LiveSync and Periodic), the liveSync-only forced teardown on becoming visible, and the default / On-Events / non-desktop paths. This module layer was previously untested because unit tests stub `obsidian`; the spec mocks `deps.ts` bare (the module needs no runtime value from it now) and drives the platform through the `services.API.isMobile()` mock, cleaning up the `activeWindow` global in `afterEach`.
- `src/lib`: submodule bump to commonlib main (`e98d929`, the #51 merge).

`docs/settings_ja.md` and `updates.md` are left to you (translation and release-note stamping). i18n needs no message-catalog change, since the setting name and description follow the existing literal-string `SettingInformation` convention.

## Testing

Verified on Windows 11 (26200), Obsidian 1.12.7, LiveSync mode, CouchDB remote, two desktop clients, measured objectively by reading the local vault file on disk:

- Background sync (setting ON, minimised): a remote edit landed locally in ~3s with the window minimised. With the setting OFF the same edit did not arrive until focus (~1m40s later), reproducing the stall.
- Stalled-channel recovery (setting ON): with the live channel blocked at the network layer for ~2.5 min (past the 30s heartbeat, so the socket went half-open), a remote edit made during the block did not self-heal while hidden and then landed within ~1s of focusing, confirming the forced teardown-on-visible re-establishes the channel.

The code path is platform-identical (single `services.API.isMobile()` gate, no OS branching). Periodic-mode behaviour is asserted from code and the unit tests. A manual Periodic and macOS pass are easy to add if you want them on record.

## Safety / scope

- Default (setting off) behaviour is unchanged apart from one guarded early-return.
- Mobile is unaffected (`services.API.isMobile()` runtime gate plus UI hidden on mobile).
- File-event commit and shutdown/unload paths are untouched.
- No protocol or data-format change; interoperates with stock clients on the same database.
- The setting key is excluded from setup-URI / QR encoding, so it is never pushed to other (possibly mobile) devices.

Passes the repo's prettier, eslint and tsc on the changed files.

## Open questions

1. Setting name and placement (currently Sync pane, shown for LiveSync and Periodic).
2. Any objection to forcing the teardown on becoming visible, or would you gate it differently?
3. Want the `processSync` `finally` hardening as a follow-up in this PR, or tracked separately?

---

This PR was drafted by Claude Code, directed and reviewed by me (Miguel Ferreira).

### DIFF
--- a/src/modules/essentialObsidian/ModuleObsidianEvents.ts
+++ b/src/modules/essentialObsidian/ModuleObsidianEvents.ts
@@ -2,7 +2,7 @@ import { AbstractObsidianModule } from "../AbstractObsidianModule.ts";
 import { EVENT_FILE_RENAMED, EVENT_LEAF_ACTIVE_CHANGED, eventHub } from "../../common/events.js";
 import { LOG_LEVEL_NOTICE, LOG_LEVEL_VERBOSE } from "octagonal-wheels/common/logger";
 import { scheduleTask } from "octagonal-wheels/concurrency/task";
-import { Platform, type TFile } from "../../deps.ts";
+import type { TFile } from "../../deps.ts";
 import { fireAndForget } from "octagonal-wheels/promises";
 import { type FilePathWithPrefix } from "../../lib/src/common/types.ts";
 import { reactive, reactiveSource, type ReactiveSource } from "octagonal-wheels/dataobject/reactive";
@@ -146,7 +146,7 @@ export class ModuleObsidianEvents extends AbstractObsidianModule {
         const keepActiveInBackground =
             this.settings.keepReplicationActiveInBackground &&
             (this.settings.liveSync || this.settings.periodicReplication) &&
-            Platform.isDesktopApp;
+            !this.services.API.isMobile();
 
         if (isHidden) {
             if (!keepActiveInBackground) await this.services.appLifecycle.onSuspending();
--- a/src/modules/essentialObsidian/ModuleObsidianEvents.ts
+++ b/src/modules/essentialObsidian/ModuleObsidianEvents.ts
@@ -2,7 +2,7 @@ import { AbstractObsidianModule } from "../AbstractObsidianModule.ts";
 import { EVENT_FILE_RENAMED, EVENT_LEAF_ACTIVE_CHANGED, eventHub } from "../../common/events.js";
 import { LOG_LEVEL_NOTICE, LOG_LEVEL_VERBOSE } from "octagonal-wheels/common/logger";
 import { scheduleTask } from "octagonal-wheels/concurrency/task";
-import { type TFile } from "../../deps.ts";
+import { Platform, type TFile } from "../../deps.ts";
 import { fireAndForget } from "octagonal-wheels/promises";
 import { type FilePathWithPrefix } from "../../lib/src/common/types.ts";
 import { reactive, reactiveSource, type ReactiveSource } from "octagonal-wheels/dataobject/reactive";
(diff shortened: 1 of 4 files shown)


vrtmrz:
Thank you for the pull request! I think it is really good.
Just one thing: I think it would be less of a hassle to use `isMobile` or `getPlatform` from `APIService` to determine the platform.
https://github.com/vrtmrz/obsidian-livesync/blob/0e04e7d31d50a2deea6cddb81da7a0635929bfc8/src/modules/services/ObsidianAPIService.ts#L72-L95

I found out about this `isMobile` feature on their Discord quite a while ago, and it is still working fine. If it ever stops working, I will emulate it.

vrtmrz:
By the way, this `getPlatform` function looks rather dodgy too, so feel free to rewrite it, e.g., combinations such as Target Triple (or collect all truhy like desktop-macos-safari). This is used for displaying the peer platform on the P2P replicator. I just wanted the name on the order of those if statements; said, isMobile may be a safer bet.

src/modules/essentialObsidian/ModuleObsidianEvents.ts
```diff
@@ -138,12 +138,38 @@ export class ModuleObsidianEvents extends AbstractObsidianModule {
...
 
+        // Desktop opt-in (LiveSync/Periodic only): keep the background channel running while the
+        // window is hidden, instead of suspending on hide. On hide we skip the suspend for both
+        // modes (LiveSync's continuous replication and Periodic's timer both stall otherwise);
+        // becoming visible reopens normally, and for LiveSync additionally forces a teardown first
+        // (see the resume branch) so a stalled continuous channel is always replaced.
+        const keepActiveInBackground =
+            this.settings.keepReplicationActiveInBackground &&
+            (this.settings.liveSync || this.settings.periodicReplication) &&
+            Platform.isDesktopApp;
```
vrtmrz: As [mentioned](https://github.com/vrtmrz/obsidian-livesync/pull/949#issuecomment-4658976014), address here, please!

vrtmrz (changes requested):
As [mentioned](https://github.com/vrtmrz/obsidian-livesync/pull/949#issuecomment-4658976014), address here, please!

migsferro:
Done. Both `Platform.isDesktopApp` checks this PR introduced now go through `services.API.isMobile()`: the runtime guard in `ModuleObsidianEvents` and the toggle gate in `PaneSyncSettings`, matching the existing call in `ObsidianLiveSyncSettingTab`. The `TFile` import became type-only as a side effect, so `deps.ts` is no longer pulled into that module at runtime.

I kept this to `isMobile()` and left `getPlatform()` as it is, since the feature only needs the desktop/mobile distinction.

Also repointed `src/lib` to commonlib main (`e98d929`) now that #51 is merged.

vrtmrz (approved):
Thank you for the amendment!
I think this is perfect. I am so happy to merge your PR!!

PR #978, opened 2026-06-24
Remember Simple Fetch choices during interrupted initialisation

### LINKED ISSUE #977: Fast Fetch initialisation should survive transient interruptions without repeated prompts
## Summary

Fast Fetch / Simple Fetch can now resume from a persisted CouchDB sequence, but the initialisation flow is still difficult to complete on Android with a large vault. There are two remaining issues: transient interruptions fail the initialisation process immediately, and the same Simple Fetch choices are asked again when retrying the same pending Fetch All operation.

## Environment

- Remote type: CouchDB
- Client: Android Obsidian
- Vault size: large existing remote database
- Remote database size observed on the server: about 5.4 GB after compaction
- Document count observed on the server: roughly 146k documents
- Network: slow VPS/mobile connection, around 0.5-0.7 MiB/s observed during earlier diagnosis
- Version tested: v0.25.78 or later, with the Fast Fetch resume fix included

### 1. Simple Fetch choices are asked repeatedly

During scheduled Fetch All / Simple Fetch, the plug-in first asks the user to choose one of:

- \`Compare time and take newer\`
- \`Overwrite all with remote files\`
- \`Use the detailed flow\`
- \`Cancel\`

After choosing either \`Overwrite all with remote files\` or \`Compare time and take newer\`, the plug-in asks a second deletion-handling question. The exact choices depend on the first choice.

For \`Overwrite all with remote files\`:

- \`Delete local files if not on remote\`
- \`Keep local files even if not on remote\`

For \`Compare time and take newer\`:

- \`Delete local files if deleted on remote\`
- \`Keep local files even if deleted on remote\`

Both flows also include:

- \`Cancel all and reboot\`

If the initialisation is interrupted or fails, the next launch asks the same choices again, even though the user already made the same decision for the same pending Fetch All operation.

### 2. Transient network or background interruptions fail initialisation

The new resume checkpoint works after a full restart. I can see logs like:

\`Resuming fast database fetch from sequence: ...\`

However, if Obsidian is backgrounded on Android and then brought back, or if the network changes from Wi-Fi to mobile data, the initialisation can stop with:

\`Error during vault initialisation process.\`

At that point, the user has to restart or re-enter the flow manually.

## Expected behaviour

- Once the user chooses the Simple Fetch mode and deletion handling for a pending Fetch All operation, the choice should be remembered locally.
- The choice should be kept until that operation completes, is explicitly cancelled, or is finalised.
- A transient fetch failure during Fast Fetch should retry a few times with exponential backoff.
- Each retry should use the latest persisted Fast Fetch checkpoint, so it continues from the last successfully written CouchDB sequence.
- The user should not be asked the same Simple Fetch choices again just because the previous attempt was interrupted.
- If all retry attempts fail, the current checkpoint should be kept so the next launch can still resume.

## Notes

This is a follow-up to #972 / #973. The original fix made Fast Fetch resumable after restarting Obsidian. This issue is about making the same initialisation flow more tolerant of transient interruptions and less repetitive for the user.

## Locally tested fix direction

- Store the Simple Fetch stage1/stage2 choices in local small config for the same pending Fetch All operation.
- Clear the stored choice after successful initialisation, explicit cancellation, or finalisation; keep it after ordinary interruption or failure so the next launch can reuse it.
- Retry transient Fast Fetch failures up to 4 times, with 2s, 5s, 10s, and 20s delays.
- Before each retry, read the latest persisted checkpoint and continue from the last successfully written sequence.
- Treat only the plug-in's own target-sequence abort as successful completion; external network or system \`AbortError\`s are propagated to the retry layer.

apple-ouyang:
## Summary

Remember pending Simple Fetch choices during interrupted initialisation, and update `src/lib` for Fast Fetch retry handling.

## Context

- Fixes #977
- Companion commonlib PR: https://github.com/vrtmrz/livesync-commonlib/pull/59

## Changes

- Store the Simple Fetch stage1/stage2 choices in local small config for the same pending Fetch All operation.
- Reuse the stored choice after ordinary interruption or failure, so the user is not asked the same setup questions again.
- Clear the stored choice after successful completion, explicit cancellation, or finalisation.
- Update `src/lib` to the commonlib commit that retries transient Fast Fetch interruptions from the latest checkpoint.

## Verification

- `npm test -- --config vitest.config.unit.ts src/serviceFeatures/redFlag.unit.spec.ts src/lib/src/pouchdb/StreamingFetch.unit.spec.ts src/lib/src/serviceModules/Rebuilder.unit.spec.ts`
- `npm run tsc-check`
- `npx eslint --no-warn-ignored src/serviceFeatures/redFlag.simpleFetch.ts src/serviceFeatures/redFlag.unit.spec.ts src/lib/src/pouchdb/StreamingFetch.ts src/lib/src/pouchdb/StreamingFetch.unit.spec.ts src/lib/src/serviceModules/Rebuilder.ts src/lib/src/serviceModules/Rebuilder.unit.spec.ts`
- `npm run build`

`npm run build` passes with existing Svelte warnings about referenced initial values.

### DIFF
--- a/src/serviceFeatures/redFlag.simpleFetch.ts
+++ b/src/serviceFeatures/redFlag.simpleFetch.ts
@@ -24,9 +24,73 @@ export const SIMPLE_FETCH_STAGE2_NEWER_CLEANUP = "Delete local files if deleted 
 export const SIMPLE_FETCH_STAGE2_NEWER_SYNC_ALL = "Keep local files even if deleted on remote";
 export const STAGE2_ABORT = "Cancel all and reboot";
 
+const SIMPLE_FETCH_MODE_KEY = "simple-fetch-mode";
+
+function buildSimpleFetchResult(stage1: string, stage2?: string) {
+    if (stage1 === SIMPLE_FETCH_STAGE1_LEGACY) {
+        return { mode: "legacy", options: {} };
+    }
+    if (stage1 === SIMPLE_FETCH_STAGE1_REMOTE_WINS && stage2) {
+        if (![SIMPLE_FETCH_STAGE2_REMOTE_DELETE_ALL, SIMPLE_FETCH_STAGE2_REMOTE_DELETE_NONE].includes(stage2)) {
+            return undefined;
+        }
+        return {
+            mode: "remote-only",
+            options: {
+                mode: FullScanModes.DB_APPLY,
+                extraOnRemote:
+                    stage2 === SIMPLE_FETCH_STAGE2_REMOTE_DELETE_ALL ? ExtraOnRemote.DELETE_LOCAL_MISSING : undefined,
+            },
+        };
+    }
+    if (stage1 === SIMPLE_FETCH_STAGE1_NEWER_WINS && stage2) {
+        if (![SIMPLE_FETCH_STAGE2_NEWER_CLEANUP, SIMPLE_FETCH_STAGE2_NEWER_SYNC_ALL].includes(stage2)) {
+            return undefined;
+        }
+        return {
+            mode: "newer-wins",
+            options: {
+                mode: FullScanModes.NEWER_WINS,
+                extraOnLocal:
+                    stage2 === SIMPLE_FETCH_STAGE2_NEWER_CLEANUP
+                        ? ExtraOnLocal.DELETE_DB_DELETED
+                        : ExtraOnLocal.APPEND_STORAGE_ONLY,
+            },
+        };
+    }
+    return undefined;
+}
+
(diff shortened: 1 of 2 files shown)


vrtmrz:
I verified this against the PR with `src/lib` fixed to the current commonlib (merged main) and `octagonal-wheels` updated to `0.1.47`.

The targeted maintenance prerequisite unit test, `tsc --noEmit`, Prettier check, ESLint check, and the full unit test suite all passed. ESLint only reported the existing ignored-spec-file warning.
This is ready to merge, with the dependency/submodule pointers to the current commonlib main. Thank you for your contribution!

PR #981, opened 2026-06-25
Ask before applying maintenance prerequisite settings

### LINKED ISSUE #980: Maintenance actions only show notices for required chunk settings
## Problem

Local database maintenance actions require specific chunk settings before they can safely inspect or delete chunks:

- `Compute revisions for chunks` must be enabled.
- `Fetch chunks on demand` / `Read chunks online` must be disabled.

Currently, when these prerequisites are not met, the maintenance command only shows a notice and exits. For example, Garbage Collection reports that `Read chunks online` should be disabled, but the setting is hidden behind Advanced mode, so users can get stuck at the maintenance action without a clear path to continue.

## Why this matters

Garbage Collection and related maintenance operations are already destructive or high-impact actions. If a required setting can be safely changed to the required state, the command should surface that dependency at the point of use and ask for confirmation instead of forcing users to hunt through settings.

## Suggested behavior

When a local database maintenance action is invoked and the required settings are not already satisfied, show a confirmation dialog such as:

```text
Garbage Collection requires the following settings:

- Compute revisions for chunks: On
- Fetch chunks on demand: Off

Apply these settings and continue?
```

If the user confirms, apply the required settings and continue with the existing maintenance flow. If the user cancels, stop without changing settings.

This should not replace existing safety confirmations around device progress, missing node information, or remote locking. It only covers the settings prerequisites that currently cause an immediate notice-and-exit path.

apple-ouyang:
## Summary
- prompt before local database maintenance actions when required chunk settings are disabled
- let users apply Compute revisions for chunks and disable Fetch chunks on demand, then continue the selected action
- cover the prerequisite flow with focused unit tests

Fixes #980

## Verification
- npm run test:unit -- src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.unit.spec.ts
- npm run tsc-check
- npx prettier --config ./.prettierrc.mjs --check src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.ts src/features/LocalDatabaseMainte/maintenancePrerequisites.ts src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.unit.spec.ts
- npx eslint src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.ts src/features/LocalDatabaseMainte/maintenancePrerequisites.ts src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.unit.spec.ts

### DIFF
--- a/src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.ts
+++ b/src/features/LocalDatabaseMainte/CmdLocalDatabaseMainte.ts
@@ -18,6 +18,7 @@ import { EVENT_ANALYSE_DB_USAGE, EVENT_REQUEST_PERFORM_GC_V3, eventHub } from "@
 import type { LiveSyncCouchDBReplicator } from "@lib/replication/couchdb/LiveSyncReplicator";
 import { delay } from "@lib/common/utils";
 import { isNotFoundError } from "@lib/common/utils.doc";
+import { ensureLocalDatabaseMaintenancePrerequisites } from "./maintenancePrerequisites";
 // import { _requestToCouchDB } from "@/common/utils";
 const DB_KEY_SEQ = "gc-seq";
 const DB_KEY_CHUNK_SET = "chunk-set";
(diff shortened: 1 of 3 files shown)


vrtmrz:
Thank you for the PR! I verified this against the PR with current `src/lib`. 
With `octagonal-wheels` updated to `0.1.47`, the targeted Simple Fetch/Fast Fetch tests, `tsc --noEmit`, and the production build all passed. This is ready to merge!

PR #991, opened 2026-07-03
docs: add managed CouchDB hosting option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we host a managed CouchDB sync server that Self-hosted LiveSync can connect to, so this PR adds a short note about it under How to Use (https://zenith.hosting/host/obsidian-livesync).

updated per your review: the note is now explicitly framed as a third-party hosting option proposed by Zenith Hosting, and it states plainly that it is not an official Self-hosted LiveSync service and is neither endorsed nor recommended by the project. i also added a line telling readers their data will sit on a server run by a third party, since that seemed like the honest thing to put next to it.

understood on sponsorships, thanks for being upfront about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,13 @@ Each workflow establishes ordinary note synchronisation on the first device, gen
 > CouchDB can also be run on a Raspberry Pi (please be mindful of your server's security).
 
 
+### Third-party managed CouchDB hosting
+
+> [!NOTE]
+> The following is a third-party hosting option proposed by Zenith Hosting. It is not an official Self-hosted LiveSync service, and it is neither endorsed nor recommended by this project.
+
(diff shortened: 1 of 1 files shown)


vrtmrz:
Thank you for the proposal. I am open to listing useful third-party hosting options, but I would like the documentation to make the relationship clear for users.

Could you please update the wording so that this is explicitly described as a third-party hosting option proposed by Zenith Hosting, not an official Self-hosted LiveSync service or an endorsed/recommended provider?

(This is only for transparency) Also, whilst I appreciate your offer, I only accept explicit sponsorships as vrtmrz.

odpay:
thanks, that is a fair ask. i have rewritten the note so it is explicit that this is a third-party hosting option proposed by Zenith Hosting, not an official Self-hosted LiveSync service and not endorsed or recommended by the project. i also added a sentence noting the data will reside on a server operated by a third party, so readers can weigh that up themselves.

it is now its own "Third-party managed CouchDB hosting" section with a NOTE callout rather than sitting inline with the setup guidance, which i hope keeps the separation clear. happy to adjust the wording further if it still reads too close to an endorsement.

noted on sponsorships, thanks for saying so directly.

vrtmrz:
Sorry to keep you waiting. And, thank you for your amendment!
I believe this has enriched users’ choices.
I am delighted to merge it!

Issue #993, opened 2026-07-04
[BUG] Conflict resolution mixes deleted old content with newly added content


SeleiXi:
I encountered a conflict case where deleted content and newly added content were both included in the conflict result.

# Scenario

1. The note had some existing content synced between desktop and mobile.
2. On desktop, I deleted some old content.
3. On mobile, before or during sync, I added new content to the same note.
4. After syncing, LiveSync detected a conflict.

# Actual behavior

The conflict result contained both:

* the old content that had already been deleted on desktop
* the new content added on mobile

Both were grouped into the conflict content, so the final conflict note showed old deleted content and newly added content together.

# Expected behavior

The deleted content and newly added content should be handled separately, or the conflict result should make it clear which part was deleted and which part was newly added.

Ideally, if one side only deleted old content and the other side only added new content, LiveSync should preserve the new content without reintroducing the deleted old content, unless both edits truly conflict on the same lines.

SeleiXi:
A possible fix might be to treat “deleted on one side, unchanged on the other side” as a safe deletion

vrtmrz:
Thank you for opening the issue!
I have reproduced it in the test code! (Well, it means the work on the infrastructure has finally started to pay off).

This will be fixed in the next version!

As the behaviour has become a bit complicated, I have jotted down a brief explanation in the documentation.

vrtmrz:
Thank you again for reporting this, and I am sorry for the delay in following up.

The Markdown conflict merge did not correctly distinguish between a safe deletion and an overlapping delete-and-edit operation. It could therefore reintroduce content that had already been deleted.

This was fixed in v0.25.80. Non-overlapping deletions and edits are now merged, while a deletion and modification of the same line remain visible as a conflict.

Would you mind checking whether your original conflict scenario now behaves as expected, please?

SeleiXi:
> Thank you again for reporting this, and I am sorry for the delay in following up.
> 
> The Markdown conflict merge did not correctly distinguish between a safe deletion and an overlapping delete-and-edit operation. It could therefore reintroduce content that had already been deleted.
> 
> This was fixed in v0.25.80. Non-overlapping deletions and edits are now merged, while a deletion and modification of the same line remain visible as a conflict.
> 
> Would you mind checking whether your original conflict scenario now behaves as expected, please?

Thank you! I will test that soon

SeleiXi:
> Thank you again for reporting this, and I am sorry for the delay in following up.
> 
[...]

Issue #795, opened 2026-02-05
Document format requirements for API/external integration


GuideThomas:
## Summary

 When integrating external services (like Slack bots, automation tools) with LiveSync via the CouchDB API, there are critical format requirements that are not documented. After debugging sync issues for hours, I discovered these rules:

 ## Critical Requirements

 ### 1. Document `_id` MUST be lowercase

 ```javascript
 // ❌ WRONG - Document won't appear in Obsidian clients
 {
   "_id": "My-Note.md",
   "path": "My-Note.md"
 }

 // ✅ CORRECT
 {
   "_id": "my-note.md",        // lowercase!
   "path": "My-Note.md"         // original case preserved here
 }
 ```

 ### 2. Folder paths in `_id` must also be lowercase

 ```javascript
 // ❌ WRONG
 { "_id": "Folder/SubFolder/Note.md" }

 // ✅ CORRECT
 {
   "_id": "folder/subfolder/note.md",
   "path": "Folder/SubFolder/Note.md"
 }
 ```

 ### 3. Chunk IDs must be alphanumeric only

 ```javascript
 // ❌ WRONG - URL encoding issues
 { "_id": "h:+abc123xyz" }    // + becomes space or %2B
 { "_id": "h:abc/123" }       // / breaks URL

 // ✅ CORRECT
 { "_id": "h:abc123xyz789" }  // alphanumeric only
 ```

 ## Complete Document Structure

 ```javascript
 // Main document
 {
   "_id": "folder/my-note.md",           // lowercase path
   "children": ["h:abc123xyz"],          // alphanumeric chunk IDs
   "path": "Folder/My-Note.md",          // original path
   "ctime": 1707123456789,
   "mtime": 1707123456789,
   "size": 42,
   "type": "plain",
   "eden": {}
 }

 // Chunk document
 {
   "_id": "h:abc123xyz",
   "data": "# Note content here",
   "type": "leaf"
 }
 ```

 ## Symptoms of Incorrect Format

 - Document exists in CouchDB but doesn't appear in Obsidian
 - "Failed to read file: Possibly unprocessed or missing"
 - "No chunks were found for the following IDs: h:+..."

 ## Environment

 - Obsidian LiveSync: Latest
 - CouchDB: 3.5.1
 - LiveSync Bridge: Latest

shorpenlee123-cyber:
[...]

PR #1001, opened 2026-07-08
Add CouchDB API compatibility spec for external integrations

oleduc:
## Summary

Adds `docs/couchdb-api-compatibility-spec.md`, a comprehensive technical specification for building external software (bots, automation tools, integrations) that reads from or writes to a LiveSync-managed CouchDB database. Motivated by [vrtmrz/obsidian-livesync#795](https://github.com/vrtmrz/obsidian-livesync/issues/795), where external-integration authors hit undocumented format requirements (lowercase IDs, chunk-ID character rules) and had open questions about how deletes propagate.

The spec is derived from source analysis of the plugin and the `livesync-commonlib` type declarations, and is aligned with plugin **v0.25.80** (protocol `VER = 12`).

## Contents

- **Document types & ID conventions** — the full `EntryTypes` set, and the ID prefix map from commonlib constants: `h:` chunks (`h:+` when E2EE), `f:` obfuscated paths, `i:` internal files, `ix:` customization sync, `ps:` plugin settings. Clarifies that `"plain"` = text vs `"newnote"` = binary (not new-vs-legacy), with `"notes"` the legacy inline shape.
- **Note & chunk (leaf) schemas** — required/optional fields, the non-optional `eden` field, and the write-chunks-before-note ordering rule.
- **Chunking** — hash algorithms (`hashAlg`), content-splitter versions, size constants, safe chunk-ID generation.
- **Delete operations** — the two mechanisms (`deleted` soft-delete vs `_deleted` tombstone) and the recommended procedure for external writers.
- **Conflict resolution** — documents the v0.25.80 conservative three-way merge policy (#993/#994): non-overlapping edits merge, honored deletions aren't resurrected, delete-vs-modify becomes a user-facing conflict, and an incoming entry only overwrites local storage without conflict when it clearly extends the local text. Includes the practical implications for external writers (append-style writes are conflict-friendly; newest-mtime does not win).
- **Sync protocol, special/system documents, E2EE & path obfuscation, protocol constants, common failure modes, and minimal working examples.**

## Notes

- Documentation only — no code changes.


oleduc:
Hello @vrtmrz is there anything that needs to change to get this merged? Thank you for your time.

vrtmrz:
Thank you for the PR; I apologise for taking up your time.
I am currently looking into whether the Project documentation itself was correct. Please bear with me for just a little longer.

vrtmrz:
Thank you very much for your patience and for this substantial contribution.

Reviewing this PR revealed that the project's own data-structure documentation had fallen behind the current implementation and contained several inaccurate statements. Your investigation was what brought this gap to light. I have now merged #1160 to correct that project-owned foundation for Self-hosted LiveSync 1.0.24 and `@vrtmrz/livesync-commonlib` 0.1.21.

I still think the purpose of this PR is valuable and worth landing. Before merging it, however, the external compatibility specification needs to be updated against the corrected baseline. In particular:

- current `xxhash64` Chunk identifiers use a base-36 hash of a length-qualified piece, rather than a hexadecimal digest;
- Metadata `size` is the decoded byte length, rather than JavaScript `String.length`;
- file-name case folding depends on the configured case-handling setting;
- `type: "plugin"` is not the current Customisation Sync format;
- logical deletion normally preserves the existing Metadata fields and Chunk references, while `_deleted: true` is a CouchDB tombstone;
- the historical fixed 5-second and 30-second waits no longer describe current missing-Chunk handling; and
- compression, E2EE, and property or path obfuscation can all change the raw remote representation.

Would you be willing to bring this PR up to date with current `main`, which now includes #1160, and revise it around an explicitly declared support matrix? It is entirely reasonable to begin with a deliberately narrow contract, for example ordinary text and binary files on CouchDB with E2EE, compression, and path obfuscation disabled, while marking other combinations as unsupported or outside the initial scope.

For the configurations presented as supported, I would also like the examples to be backed by an executable CouchDB compatibility fixture covering create, read, update, logical deletion, Unicode byte sizes, stale revisions, and missing Chunks.

If that makes the contribution larger than you intended, please let me know. I would be very happy to help narrow or divide the remaining work while preserving the external-integration specification you have already prepared.

oleduc:
@vrtmrz Yes I will take care of it. Thank you for your time!

Issue #1006, opened 2026-07-13
Repeated ignore matching retains oversized minimatch WeakMap tables


apple-ouyang:
### Description

Long-running vault scans can leave the Obsidian renderer with hundreds of megabytes retained by `minimatch` AST private-field `WeakMap` tables.

### Environment

- Obsidian 1.12.7 on macOS
- Self-hosted LiveSync 0.25.80
- Vault with about 60,000 files
- Plugin process had been running for about 18 hours

### Heap snapshot evidence

A V8 heap snapshot reported:

- 1,165.92 MB total JavaScript self size
- twelve 64 MB arrays retained as the internal tables of eleven `WeakMap`s and one `WeakSet`
- 768 MB total in those tables, about 66% of the JavaScript heap
- the retaining context variables were `_root`, `_hasMagic`, `_uflag`, `_parts`, `_parent`, `_parentIndex`, `_negs`, `_filledNegs`, `_options`, `_toString`, `_emptyExt`, and `_AST_instances`
- these symbols resolve to the `minimatch` AST implementation bundled in LiveSync
- each 64 MB table had only 4,399 snapshot edges, consistent with a high-water backing table left after repeated AST creation

The raw heap snapshot is not attached because heap snapshots can contain vault contents, paths, configuration, and credentials.

### Root cause

`livesync-commonlib/src/string_and_binary/path.ts::isAccepted()` calls the top-level `minimatch(path, pattern, options)` function for every checked path and ignore pattern. The top-level function constructs a new `Minimatch` and parses the pattern on every call.

Ignore file contents are already cached by `isAcceptedByIgnoreFilesFactory()`, but their compiled matchers are not. Repeated vault scans therefore create the same `minimatch` ASTs again for stable ignore arrays.

### Expected behavior

Stable ignore arrays should reuse compiled `Minimatch` instances. Updating or invalidating an ignore file should allow its old matchers to be garbage collected.

### Proposed fix

Cache compiled matchers by ignore-array identity with a `WeakMap<string[], Map<string, Minimatch>>`. This preserves the existing API and matching semantics while avoiding a global string cache that would retain historical patterns indefinitely.

apple-ouyang:
I completed a second pass over the same heap snapshot and the repository's remaining `minimatch` call sites.

After excluding the twelve oversized minimatch tables, the next largest individual allocations were:

[...]

PR #1007, opened 2026-07-13
Compile CLI ignore patterns once

apple-ouyang:
## Summary

- compile CLI ignore patterns when rules are loaded
- reuse `Minimatch` instances for every subsequent path check
- add a regression test that counts matcher construction

## Why

The long-running CLI daemon used the top-level `minimatch()` function in `shouldIgnore()`. Like the Obsidian-side path filter reported in #1006, this reparsed stable glob patterns for every checked path and could produce the same minimatch AST churn.

This change stores compiled matchers directly in `IgnoreRules.patterns`. Calling `load()` still replaces the full ruleset, so old matchers are released without adding another cache.

Refs #1006

## Verification

- `vitest run --config vitest.config.unit.ts`: 411 suites, 1369 tests passed
- `npm run tsc-check`
- `npm run lint` (one unrelated existing warning)
- `npm run build` (existing Svelte warnings only)

### DIFF
--- a/src/apps/cli/serviceModules/IgnoreRules.ts
+++ b/src/apps/cli/serviceModules/IgnoreRules.ts
@@ -1,4 +1,4 @@
-import { minimatch } from "minimatch";
+import { Minimatch } from "minimatch";
 import { fsPromises as fs, path } from "@/apps/cli/node-compat";
 
 /**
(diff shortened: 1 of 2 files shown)


vrtmrz:
Thank you for your pull request! This approach looks sound to me.
The equivalent code path in the main plug-in will still need to be addressed separately under #1006, but this should resolve the problem for the CLI. I am happy to merge your contribution!

PR #1009, opened 2026-07-13
feat: Add revision prev/next buttons next to history slider

### LINKED ISSUE #990: feat: easier revision navigation in Document History
I would like to suggest a small UX improvement for the Document History modal.

Currently, when reviewing a note’s history, the revision can be changed through the range slider. This works, but it is sometimes hard to move exactly one revision backward or forward, especially when there are many revisions.

Two options I'm considering:

**A. Reuse the existing ◀ / ▶ in the search row (dual-mode)**  
- Empty search → step older/newer revision  
- Active search → jump between matches (current behavior)  
- Pros: no extra controls; one consistent nav affordance  
- Cons: behavior changes with context; might be less discoverable

**B. Add ◀ / ▶ next to the range slider**  
- Search row buttons stay search-only  
- Slider row gets dedicated revision prev/next  
- Pros: clearer separation of concerns  
- Cons: more buttons; slider row gets busier

Do you have a preference, or another pattern you'd recommend for Document History?

SeleiXi:
close #990 

 ## Summary

  Adds dedicated previous/next revision buttons beside the Document History revision slider.

  ## Changes

  - Add ◀ / ▶ buttons next to the history slider.
  - Show the current revision position with a `Rev x/y` indicator.
  - Disable revision navigation buttons at the oldest/newest boundaries.
  - Keep existing search-result navigation separate from revision navigation.
  - Add an Obsidian E2E script for validating revision navigation in the Document History modal.

  ## Verification

  - Built the plugin with `npm run build`.
  - Ran a real Obsidian UI test against the migrated branch.
  - Verified:
    - initial state showed `Rev 3/3`
    - clicking ◀ moved to `Rev 2/3`
    - clicking ▶ returned to `Rev 3/3`
    - search history UI still reported matches correctly
01-initial-latest-rev
image


vrtmrz:
Thank you for this improvement. I tested it against v1.0.0. This PR is conflict-free, and `npm run check` completes for this PR.

I found three integration points which should be addressed before merging:

1. Ordinary ESLint reports a new `obsidianmd/prefer-create-el` warning for the revision indicator. Please use `revNavRow.createSpan({ text: "\u2014" }, ...)` instead of
`createEl("span", ...)`.
2. `git diff --check` reports a new blank line at the end of `test/e2e-obsidian/scripts/document-history-nav.ts`.
3. The new real-Obsidian scenario is not registered as a package script or supported focused scenario. When invoked directly, it also times out against the current 1.0 start-up contract: a fresh unconfigured Vault intentionally remains suspended, so the first `waitForLiveSyncCoreReady()` cannot complete. Applying `isConfigured` after that start does not make the already-started
lifecycle ready either.

For the third point, please preseed the local test settings through the `pluginData` option of `startObsidianLiveSyncSession`, following the existing `revision-repair.ts` or `conflict-dialog-
policy.ts` pattern, and then wait for core readiness. 

I think I have changed the behaviour, but it should now be more robust. I am sorry for the extra trouble of having to read through the code.

vrtmrz:
Thank you for your refinement! It works great in the small UIs!

Issue #130, opened 2022-10-11
Folder rename case not respected


Vyeche:
Hi, 

Thank you for the plugin. Loving it! :) 

Renaming the case of the folder isn't respected. So, I have to rename it for the sync to work manually.
- InputNotes -> inputNotes
- I -> i

![image](https://user-images.githubusercontent.com/972377/195114139-f1303456-8d4e-4ff2-a7bd-9cb1551197ea.png)
![image](https://user-images.githubusercontent.com/972377/195114574-1e90c39b-6d10-4c14-a9ad-212ac701f9ce.png)

Please advise.

vrtmrz:
Thank you for finding the edge case. 
I have a bad feeling about this. We have to investigate it more, but in renaming, saving the new file is prior to deleting it by the design. (Just I wanted not to lose files even when hanged). Then, possibly we are going to lose the file. wow...

I didn't reproduce it but I thought I should tell you this.

(Of course, we can roll it back in the `History`)

Could you enable the verbose logging and check the detail of the error following the line `Could not parse`?
In the current plan, it will be fixed by changing the processing order.

vrtmrz:
@Vyeche 
Sorry for being late. It has been fixed in case-sensitive platforms.
(Unfortunately, Windows does not respect cases. But it would not be a `Parse error` again.)

Vyeche:
Got it, I'll test it out and see if it's okay. Thanks!

Vyeche:
Hmm, not working, after rename. I tried opening my mobile device. Got an error.

it deleted the file on my mobile device. Then

vrtmrz:
For future reference, #1014 now safely handles case-only file renames when the parent directory is unchanged, for example:

`Notes/Calculus.md` → `Notes/calculus.md`

It does not implement case-only folder renames. With case-insensitive filename handling enabled, an incoming change such as:

`Notes/Calculus.md` → `notes/Calculus.md`

is now deliberately refused as a file rename. The existing file and its contents remain at the old-cased path, rather than partially moving child files or leaving Obsidian's Vault cache inconsistent. The database revision may therefore remain unapplied to the Vault until the folder casing is aligned.

A complete solution requires Self-hosted LiveSync to propagate the folder rename itself and apply it as a single directory-aware operation. Moving the child files individually is not equivalent on case-insensitive file systems such as Windows.

I am leaving this historical issue closed for now, but recording here that folder case-only renames remain outside the scope of #1014.

PR #1014, opened 2026-07-14
fix: synchronise case-only file renames safely

### LINKED ISSUE #198: `WRITE ERROR` upon changing the capitalisation of existing notes causing them to be destroyed
Changing the capitalisation of existing notes causes the file to be destroyed with a `WRITE ERROR` tooltip.

## How to replicate

1. Create a note called 'example'
2. Rename the note to 'Example'
3. The note is destroyed and a `WRITE ERROR` tooltip is displayed

## Workaround

1. Create a note called 'example'
2. Rename the note to 'example1'
3. Rename the note to 'Example'
4. The note is not destroyed

metrovoc:
## Summary

This PR addresses #198, where renaming a note by changing only its letter case (e.g. `Calculus.md` → `calculus.md`) can lead to the note being deleted on other devices when `handleFilenameCaseSensitive` is disabled: the rename is currently processed as a delete/create pair on one shared document, and the resulting tombstone can replicate out as a real deletion.

It updates `livesync-commonlib` to process renames as one atomic `RENAME` event, and implements the storage side on each local adapter:

- Bump the `src/lib` submodule and regenerate the fallback `_types` declarations (348 of the changed files are these regenerated stamps).
- Obsidian adapter: delegate renames to `app.vault.rename`.
- Node adapter: fs renames with destination-directory creation, cache refresh, and stored-case verification on case-insensitive file systems.
- File System Access API adapter: rename via a durable temporary copy (the source is restored if target creation fails), cache refresh, and per-component case verification.
- Add adapter regression tests (real temp-dir Node renames; in-memory case-insensitive FSA) and an Unreleased changelog note.

## Dependency

This depends on vrtmrz/livesync-commonlib#68 — the submodule pointer targets its head commit, so that PR would need to be merged first.

## Testing

- `npm run check`, `npm run build`
- `npm run test:unit` — 72 test files, 1,403 tests
- `npm run build` in `src/apps/cli` and `src/apps/webapp`

Fixes #198


vrtmrz:
I am sorry that I have already posted the note, which was drafted in my Obsidian, to the commonlib repo.

As I mentioned over there, I will be taking a few additional steps, but I am in favour of your approach. I will get on with it.

metrovoc:
Thanks for the quick and detailed response, and for testing it on real devices! The follow-up plan sounds good to me, and glad I could help :)

vrtmrz:
Thank you very much for getting this long-standing issue moving again. Now that the testing infrastructure is in place, I was able to validate the change in a comparatively straightforward way. I am genuinely delighted that this fundamental behaviour has finally been fixed.

While integrating the contribution, I rebased it onto the current `main` and added a guard against mistaking a file case change for a parent-directory case change, together with adapter and real-Obsidian regression coverage.

This resolves the case-only note rename scenarios reported in #198. Case-only directory renames remain a separate limitation previously discussed in #130; this PR deliberately avoids treating them as file-only renames.

The commonlib changes are now merged in vrtmrz/livesync-commonlib#68 and vrtmrz/livesync-commonlib#72. Validation passed across the focused adapter and commonlib tests, the full downstream unit suite, build and static checks, a two-vault real-Obsidian E2E run, and Windows probes for direct file and folder case changes.

Thank you again for the contribution!

metrovoc:
Glad to see it all landed smoothly. Thank you again for the thorough validation and follow-up work!

PR #1017, opened 2026-07-14
fix: suppress bulk conflict resolution notice spam

### LINKED ISSUE #1016: [BUG] Bulk conflict resolution floods mobile with per-file merge notices
### Description

Running **Resolve All conflicted files by the newer one** on a large vault creates a separate Obsidian notice for every resolved file:

```text
<path> has been merged automatically
```

With thousands of files being checked, many notices are displayed simultaneously and make the mobile UI almost unusable.

A screenshot is intentionally omitted because it contains private file names.

### Steps to reproduce

1. Open a large vault containing conflicted revisions.
2. Run **Resolve All conflicted files by the newer one**.
3. Wait while **Check and Processing** scans the vault.
4. Observe that each successfully resolved file creates a separate notice.

Observed with Self-hosted LiveSync 0.25.80 on mobile while 8,854 files were being checked. The same code is still present in 0.25.81.

### Expected behaviour

Per-file successful resolutions performed by this bulk operation should be written to the log without creating separate notices. The operation should retain its progress and completion notices, and errors should remain notices.

### Root cause

`_resolveConflictByDeletingRev()` treats the `NEWEST` bulk-resolution path as `LOG_LEVEL_NOTICE`:

https://github.com/vrtmrz/obsidian-livesync/blob/056d89aa4d1c0c53e54639b381d24776c5fc1bb9/src/modules/coreFeatures/ModuleConflictResolver.ts#L61-L62

Each message contains a different path, so the notification system uses a different default key and the notices stack.

The progress condition is also inverted:

https://github.com/vrtmrz/obsidian-livesync/blob/056d89aa4d1c0c53e54639b381d24776c5fc1bb9/src/modules/coreFeatures/ModuleConflictResolver.ts#L215-L223

`if (i++ % 10)` is true for nine out of every ten files rather than once every ten files.

### Suggested fix

- Log successful `NEWEST` resolutions at `LOG_LEVEL_INFO`, while retaining notices for errors and non-bulk conflict resolution.
- Increment the processed counter explicitly and update progress when `i % 10 === 0`.

Related but broader notification request: #770.

apple-ouyang:
## Summary

- suppress per-file success notices only during bulk newest-revision conflict resolution
- retain the existing success notice for non-bulk `resolveByNewest` calls
- update the keyed bulk progress notice once every ten checked files
- add focused unit coverage for bulk suppression, non-bulk notices, and progress updates

Closes #1016.

## Testing

- `npm run test:unit -- src/modules/coreFeatures/ModuleConflictResolver.unit.spec.ts` (3 passed)
- `npm run check` (passed; existing lint and Svelte warnings remain)
- `npm run build` (passed; existing Svelte warnings remain)
- GitHub Unit Tests, Integration Tests, and Snyk checks passed

The full local unit suite passed 1,382 of 1,383 tests. The unrelated `src/rabinKarpBom.unit.spec.ts` test failed consistently on this checkout; GitHub's full Unit Tests check passed.

The reproduction screenshot is intentionally not included because it contains private vault paths.

### DIFF
--- a/src/modules/coreFeatures/ModuleConflictResolver.ts
+++ b/src/modules/coreFeatures/ModuleConflictResolver.ts
@@ -30,7 +30,8 @@ export class ModuleConflictResolver extends AbstractModule {
     private async _resolveConflictByDeletingRev(
         path: FilePathWithPrefix,
         deleteRevision: string,
-        subTitle = ""
+        subTitle = "",
+        showNotice = true
     ): Promise<typeof MISSING_OR_ERROR | typeof AUTO_MERGED> {
         const title = `Resolving ${subTitle ? `[${subTitle}]` : ""}:`;
         if (!(await this.core.fileHandler.deleteRevisionFromDB(path, deleteRevision))) {
@@ -58,7 +59,7 @@ export class ModuleConflictResolver extends AbstractModule {
             this._log(`Could not write the resolved content to the storage: ${path}`, LOG_LEVEL_NOTICE);
             return MISSING_OR_ERROR;
         }
-        const level = subTitle.indexOf("same") !== -1 || subTitle === "NEWEST" ? LOG_LEVEL_INFO : LOG_LEVEL_NOTICE;
+        const level = subTitle.indexOf("same") !== -1 || !showNotice ? LOG_LEVEL_INFO : LOG_LEVEL_NOTICE;
         this._log(`${path} has been merged automatically`, level);
         return AUTO_MERGED;
     }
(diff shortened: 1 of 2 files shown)


vrtmrz:
Thank you for your PR!. The progress-update correction looks right.
However, one scope issue remains: treating every `NEWEST` resolution as information-level logging also suppresses the success notice for non-bulk resolveByNewest calls. Could you make the suppression specific to the bulk operation and add coverage showing that non-bulk resolution retains its notice?

apple-ouyang:
Thank you for pointing out the scope issue. I updated the implementation so that notice suppression is now explicit and limited to the bulk operation. Non-bulk `resolveByNewest` calls retain their success notice.

I also added coverage for both paths and kept the corrected ten-file progress update. The new Unit Tests, Integration Tests, and Snyk checks are passing.

vrtmrz:
Thank you for the amendment to the pull request! I think it is perfect.
I am delighted to have merged your contribution!

PR #1019, opened 2026-07-14
Fix stale Security Seed after remote rebuild

### LINKED ISSUE #1018: Stale Security Seed cache survives a remote database rebuild
### Abstract

A client process can keep using a cached Security Seed after another device rebuilds the remote CouchDB database and creates a new seed. The stale client then encrypts uploads with the old seed, mixing two encryption keys in the rebuilt database and causing `Decryption with HKDF failed` / `OperationError` on other clients.

### Expected behaviour

Each replication boundary should verify the current remote Security Seed before any pull or push starts. A client that remained open during a remote rebuild must not reuse the previous seed.

### Actually happened

`SyncParamsHandler` caches the promise returned by `fetchSyncParameters()`. The main replication preflight calls `ensurePBKDF2Salt(..., useCache: true)`, so a long-running client does not fetch `_local/obsidian_livesync_sync_parameters` again after another device recreates the database.

In a reproduced incident, server access logs showed:

1. Device A deleted and recreated the remote database, generating a new Security Seed.
2. Device B remained running. Its later replication started from sequence zero and uploaded documents without first requesting the sync-parameters local document.
3. A subsequent bulk pull returned 50 encrypted documents, immediately followed by exactly 50 HKDF decryption failures on the client.
4. The configured E2EE passphrase matched the canonical setup URI, excluding a passphrase mismatch.

### Reproducing procedure

1. Configure two devices against the same encrypted CouchDB remote.
2. Start both clients so Device B has cached the Security Seed.
3. Keep Device B running. On Device A, perform a remote overwrite/rebuild that recreates the database and generates a new Security Seed.
4. Allow Device B to replicate again.
5. Observe that Device B reuses the cached seed instead of fetching the new remote seed before replication.
6. Pull the mixed remote data from another client and observe `Decryption with HKDF failed` / `OperationError`.

### Obsidian debug info

The incident data contains private vault and server metadata, so it is not pasted publicly. The behaviour is also covered by a focused unit reproduction in the accompanying PR.

### LiveSync version

- Self-hosted LiveSync version: 0.25.81

### Report and Logs from LiveSync

Relevant redacted log signature:

```text
OperationError
[StackTrace]: _LiveSyncError: [Error Logged]
at outgoingDecryptHkDF
at async mapDoc
at async Promise.all
[CausedBy]: undefined
```

### Other information, insights and intuition

The smallest fix is to refresh the Security Seed in the existing `onBeforeReplicate` preflight by calling `ensurePBKDF2Salt(..., useCache: false)`. This preserves the existing cache within a replication run while preventing a stale process-wide value from crossing replication boundaries.

apple-ouyang:
## Summary

- refresh the remote Security Seed in the existing replication preflight instead of reusing the process cache
- add a focused regression test for the preflight cache policy
- document the fix under `Unreleased`

This prevents a long-running client from uploading documents encrypted with an obsolete seed after another device recreates the remote database.

Fixes #1018

## Verification

- `npm exec --yes --package=deno -- npm run check` (passes; one existing `no-explicit-any` warning)
- `npx vitest run --config vitest.config.unit.ts src/modules/core/ModuleReplicator.unit.spec.ts` (passes)
- `npm run build` (passes with existing Svelte warnings)
- `npm run test:unit` (1,380 tests pass, including the new regression test; the existing `src/rabinKarpBom.unit.spec.ts` U+FEFF boundary test fails on current `main`-equivalent code and is unrelated to this change)

Please run CI tests.

### DIFF
--- a/src/modules/core/ModuleReplicator.ts
+++ b/src/modules/core/ModuleReplicator.ts
@@ -49,7 +49,8 @@ async function canReplicateWithPBKDF2(
     // Showing message is false: that because be shown here. (And it is a fatal error, no way to hide it).
     // tagged as network error at beginning for error filtering with NetworkWarningStyles
     const ensureMessage = `${MARK_LOG_NETWORK_ERROR}Failed to initialise the encryption key, preventing replication.`;
-    const ensureResult = await replicator.ensurePBKDF2Salt(currentSettings, showMessage, true);
+    // A remote database rebuild replaces the Security Seed while this process may still hold the previous one.
+    const ensureResult = await replicator.ensurePBKDF2Salt(currentSettings, showMessage, false);
     if (!ensureResult) {
         errorManager.showError(ensureMessage, showMessage ? LOG_LEVEL_NOTICE : LOG_LEVEL_INFO);
         return false;
--- a/src/modules/core/ModuleReplicator.unit.spec.ts
+++ b/src/modules/core/ModuleReplicator.unit.spec.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from "vitest";
+import { ModuleReplicator } from "./ModuleReplicator";
+
+describe("ModuleReplicator", () => {
+    it("refreshes the remote Security Seed before replication", async () => {
+        const ensurePBKDF2Salt = vi.fn(async () => true);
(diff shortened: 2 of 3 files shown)


vrtmrz:
Thank you for your PR! I have not realised this completely.
This PR is quite a reasonable fix. I am so happy to merge your contribution!

PR #1030, opened 2026-07-17
Normalise tracked text files and formatter scope

vrtmrz:
## Summary

- normalise the four tracked CRLF or mixed-line-ending text files to LF without changing their non-line-ending content
- add LF attributes for `.gitignore`, `.dockerignore`, and `Dockerfile`, and remove a duplicate shell-script attribute
- format the seven TypeScript files identified by the repository's `prettyCheck` command
- exclude the separately maintained `src/lib` submodule from the root Prettier scope

## Why

`docs/setup_own_server_cn.md` was stored with CRLF line endings even though `.gitattributes` requires Markdown files to use LF. As a result, a fresh checkout reported the file as modified before any edit. The repository-wide EOL audit also found mixed or CRLF endings in `.dockerignore`, `src/apps/cli/.gitignore`, and `src/apps/cli/Dockerfile`, which did not yet have extensionless-file attributes.

The root `prettyCheck` command found seven unformatted TypeScript files in the CLI benchmark and shared Storage Adapter test code. Once the commonlib submodule was initialised, it also inspected two files owned by that separate repository. Ignoring `src/lib/**` keeps the formatter boundary consistent with the contribution rule that commonlib changes belong in its own repository.

The line-ending files have identical content after carriage returns are removed; in particular, this does not edit the Chinese translation. The patch appears larger because Git represents each normalised CRLF line as a replacement. There is no plug-in runtime or user-facing behaviour change, so no update note is required.

## Validation

- audited every tracked file with `git ls-files --eol`; no CRLF or mixed index/worktree entries remain
- compared non-line-ending SHA-256 values for all four normalised files; every value matches its pre-change content
- `npm run prettyCheck`
- `npm run tsc-check`
- `npm run lint` — no errors; one existing commonlib warning
- `npm run test:unit` — 1,504 tests passed
- focused Node and File System Access API Storage Adapter contracts — 21 tests passed
- `deno task test:benchmark-contract` — 6 tests passed


vrtmrz:
As this has been diverted from the latest.

PR #1032, opened 2026-07-17
Keep replicated document application active

### LINKED ISSUE #1031: Keep replicated-document application active after transfer completes
## Abstract

After moving many files and directories to reorganise a large vault on another device, Android can finish the remote transfer while thousands of downloaded documents remain in the replication-result queue. The existing finite-remote activity boundary then ends before those documents have been applied to the local vault. Moving Obsidian to the background invokes or permits suspension, and the queue stops draining until the application becomes visible again.

This is the mobile screen-lock case mentioned in #939, but it occurs after network replication rather than during it. The screen-awake work in #1021 protects finite remote transfer, rebuild, P2P, and chunk-fetch operations; it does not currently protect the downloaded-document application tail.

## Reproducing procedure

1. Use Self-hosted LiveSync on a desktop and an Android device with a vault containing several thousand notes.
2. On the desktop, move many files and directories to reorganise the vault, for example by moving notes from a single year directory into separate year and month directories.
3. Open Obsidian on Android and allow replication to pull the changes.
4. Observe that remote transfer completes while the `📥` unprocessed-transferred-items count remains in the thousands and the log continues to show `Processing <path> ... Done`.
5. Move Obsidian to the Android home screen.

## Expected behaviour

Best-effort screen-awake and LiveSync lifecycle protection should remain active until every downloaded document has been applied to local storage. Platform suspension may still stop hidden work, because a community plug-in cannot provide privileged Android background execution.

## Actual behaviour

The activity boundary ends when the finite remote operation settles. The remaining replication-result queue is local asynchronous work, so LiveSync can suspend while thousands of downloaded documents still need to be written.

## Root cause

`ModuleReplicator._parseReplicationResult` enqueues documents and immediately returns. `ReplicateResultProcessor` drains that queue independently with an existing ten-document semaphore. Therefore, increasing or exposing concurrency is not the missing contract: the queue already runs at a concurrency of ten. The missing contract is the lifetime between the first queued replicated document and the final local application and recovery snapshot.

## Proposed scope

- Enter one shared broad activity boundary when the replication-result queue becomes non-empty.
- Keep it active while queued or in-progress replicated documents remain.
- Persist the final empty recovery snapshot before normally releasing it.
- Release it when processing is explicitly suspended, and reacquire it on resume if work remains.
- Do not change the existing concurrency limit.
- Do not claim guaranteed Android background execution.

apple-ouyang:
## Summary

- retain one broad remote-activity boundary while downloaded documents remain queued or in progress for local application;
- settle the final empty recovery snapshot before normally releasing that boundary, and release or reacquire it around explicit processing suspension; and
- keep the existing ten-document concurrency limit unchanged.

Closes #1031.

## Why

A finite remote operation can finish after it has delivered documents into the local database but before `ReplicateResultProcessor` has applied them to the vault. With a large batch, the status bar may therefore show thousands of `📥` unprocessed transferred items after the network operation has already settled.

The activity work introduced in #1021 ends with the remote operation. On Android, moving Obsidian to the home screen can then invoke or permit suspension while the downloaded-document queue is still draining. The queue is snapshotted and resumes later, but a person may need to keep Obsidian visible for several minutes.

## Behaviour and boundaries

The first queued replicated document enters `runBoundedRemoteActivity` with the label `replicated-document-application`. Further documents and replication batches share the same boundary. It settles only when both the queue and in-progress set are empty and the final recovery snapshot has been attempted.

Explicitly suspending replication-result processing releases the boundary. Resuming reacquires it when queued work remains. Offline scans and unrelated local storage reflection are unchanged.

This remains best effort. A Screen Wake Lock does not guarantee hidden execution, and Android may still pause or terminate Obsidian. This change only prevents LiveSync from ending its own lifecycle protection before the delivered documents have been applied.

No concurrency setting is added. `ReplicateResultProcessor` already uses `Semaphore(10)`, which is higher than the proposed eight-way mobile value. Raising it without device evidence could increase IndexedDB and file-system contention, memory use, battery consumption, and thermal throttling without improving throughput.

## Verification

- Added a regression test covering two concurrent document applications sharing one boundary and retaining it until both settle. Before the implementation, it failed because the boundary was called zero times.
- `npm run test:unit`: 89 files and 1,505 tests passed.
- `npm run check`: TypeScript, ESLint, Svelte checking, and compatibility checks passed. The existing one ESLint warning and 22 Svelte warnings remain unchanged.
- `npm run build`: production build passed.
- `npm run check:compatibility`: the rebuilt `main.js` passed the iOS 15 compatibility rules.
- Regenerated the fallback declaration for `ReplicateResultProcessor`.

### DIFF
--- a/docs/adr/2026_07_bounded_remote_activity.md
+++ b/docs/adr/2026_07_bounded_remote_activity.md
@@ -37,9 +37,9 @@ The Obsidian host injects the screen wake-lock manager from the `octagonal-wheel
 
 Finite replication enters both counts only after readiness checks have succeeded and leaves them after `openReplication(..., continuous: false, ...)` settles. A successful completion has reached the latest sequence in that operation's scope and is therefore an authoritative quiescence boundary for chunk retrieval. A failed operation does not prove latest state, but can no longer deliver documents from that attempt. Failure handling runs afterwards so a mismatch or recovery dialogue does not retain the activity. This includes the direct start-up synchronisation path as well as manual, event-driven, and periodic calls through `ReplicationService`. The unbounded continuous channel does not enter either boundary, but its finite initial pull-only catch-up does; the one-shot parameter fallback chain remains inside that boundary.
 
-Delivery into the local database and application to the Obsidian Vault are deliberately separate lifetimes. A mobile device can have only a short opportunity to obtain remote data, while applying a large downloaded batch to the Vault is durable, offline-capable work which can continue or resume later. The replication-result queue and its recovery snapshot therefore remain outside `boundedRemoteActivityCount`. Finite remote activity ends when the transfer operation settles, even when the `📥` queue still contains documents awaiting Vault application.
+Delivery into the local database and application to the Obsidian Vault are deliberately separate lifetimes. Finite remote activity ends when the transfer operation settles, even when the `📥` queue still contains documents awaiting Vault application. The Obsidian host tracks that queue through a separate `boundedLocalApplicationActivityCount`, so local Vault writes do not increment either remote-activity count or keep the `📲` indicator visible.
 
-This separation also keeps the activity indicators truthful: `📲` describes a finite remote operation and must not remain active solely because local Vault writes are pending. The existing replication-result count continues to describe that local queue. If a future feature offers screen-awake protection while applying downloaded documents, it must use a separately typed local-application activity or power-policy boundary, preserve the current behaviour by default, and avoid incrementing either remote-activity count.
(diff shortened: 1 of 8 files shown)


vrtmrz:

apple-ouyang:
Rebased onto the current `main` and force-pushed. The fix now uses a separate local-application activity boundary, so queued Vault application retains Wake Lock and lifecycle protection without extending remote-activity reporting. Focused tests, the full unit suite (575 tests), local checks/build, and all GitHub checks pass.

vrtmrz:
Thank you so much for your amendment! This makes Self-hosted LiveSync more useful!

PR #1037, opened 2026-07-22
fix(cli): floor watch-mode stat timestamps to integer milliseconds

andrewleech:
### Summary

My phone's Obsidian started crash-looping on launch after I'd been running the headless CLI as a sync daemon. adb logcat pointed straight at a native `ClassCastException: Double cannot be cast to Long` inside Capacitor's `Filesystem.setTimes`: a synced document had a fractional-millisecond mtime, and the Java binding only accepts an integer `Long`.

It turns out the CLI is where the float gets in. On Linux `fs.Stats.mtimeMs`/`ctimeMs` come back with sub-millisecond precision, e.g. `1778511180024.462`. The earlier fix (3f7bb047) floored these at the scan/stat adapters, but the chokidar watch adapter was missed, and that's exactly the path the daemon runs on once it's up. So every live edit kept writing a float timestamp into the database, and any mobile client that replicated one would crash on its next launch.

This floors mtime/ctime at the watch site too, the same way the other adapters already do it, so watch-mode documents carry integer milliseconds like the rest of the mesh.

### Testing

tsc-check clean, and the CLI unit tests pass (7/7, including a new case that a fractional stat gets floored). I haven't run the CLI E2E suite locally though. 

The floor call is now hand-rolled at a handful of CLI stat-read sites. A shared `fs.Stats` to `UXStat` converter that floors once would be a nicer home for it and would make it impossible to add a new stat-read site that forgets, but that's a wider refactor across pre-existing code so I've left it as a follow-up. The deepest fix sits in commonlib's `dbToStorage`, where the timestamp is read from the DB document once for every platform, but that lives in the `src/lib` submodule.

### DIFF
--- a/src/apps/cli/managers/CLIStorageEventManagerAdapter.ts
+++ b/src/apps/cli/managers/CLIStorageEventManagerAdapter.ts
@@ -116,8 +116,11 @@ class CLIWatchAdapter implements IStorageEventWatchAdapter {
         return {
             path: path.relative(this.basePath, filePath).replace(/\\/g, "/") as FilePath,
             stat: {
-                ctime: stats?.ctimeMs ?? Date.now(),
-                mtime: stats?.mtimeMs ?? Date.now(),
+                // Floor to integer milliseconds: Linux fs.Stats.*Ms carry sub-millisecond
+                // precision, and a non-integer mtime stored in the database crashes mobile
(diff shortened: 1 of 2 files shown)


andrewleech:
Consolidated into #1039 - this is one issue (a float mtime crashing mobile) fixed at both ends, so it reads better as a single PR with the two commits. Closing in favour of that.

PR #1038, opened 2026-07-22
fix(storage): floor write-option timestamps in Obsidian adapters

andrewleech:
### Summary

Same root cause as #1037, but from the consumer side. My phone kept crash-looping on launch even after the bad timestamps were cleaned out of the remote database, because its local replica still held a document with a fractional mtime and applied it on startup before it could re-sync.

On mobile, Obsidian hands mtime/ctime straight through to Capacitor's `Filesystem.setTimes`, whose native binding casts the value to a Java `Long`. A float makes that cast throw `ClassCastException: Double cannot be cast to Long`, which takes down the whole app before it finishes loading. A float can arrive from any client that stored `fs.Stats.mtimeMs` without flooring, so I wanted a guard that doesn't rely on every writer in the mesh getting it right.

This coerces mtime/ctime to integer milliseconds at the Obsidian vault and storage adapter write boundary, which is the last point in this repo before the native call. With it in place the app launches again, then re-syncs and self-heals its local copy against a clean remote.

### Testing

tsc-check clean and the full plugin build succeeds. I also verified it on a real device: reproduced the launch crash from a float-mtime document, then confirmed the built plugin launches cleanly and stays up.

The single right choke point is `dbToStorage` in livesync-commonlib, where the timestamp is read off the DB document once for every platform. A guard there would cover the CLI and webapp too, but it's in the `src/lib` submodule so it can't go in this PR. Guarding per write method here does leave a drift risk: a future write method added to the adapter could forget the wrapper. I considered a construction-time proxy that intercepts the write family instead, but that trades explicit type-checked calls for string-matched method interception, which isn't worth it for two small adapters.

### DIFF
--- a/src/serviceModules/FileSystemAdapters/ObsidianStorageAdapter.ts
+++ b/src/serviceModules/FileSystemAdapters/ObsidianStorageAdapter.ts
@@ -2,6 +2,7 @@ import type { UXDataWriteOptions } from "@lib/common/types";
 import type { IStorageAdapter } from "@lib/serviceModules/adapters";
 import { toArrayBuffer } from "@lib/serviceModules/FileAccessBase";
 import type { Stat, App } from "obsidian";
+import { toIntegerTimestamps } from "./sanitizeWriteOptions";
 
 /**
  * Storage adapter implementation for Obsidian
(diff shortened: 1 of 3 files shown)


andrewleech:
Consolidated into #1039 - this is one issue (a float mtime crashing mobile) fixed at both ends, so it reads better as a single PR with the two commits. Closing in favour of that.

PR #1039, opened 2026-07-22
fix: prevent float mtime from crashing mobile clients

andrewleech:
### Summary

My phone's Obsidian started crash-looping on launch after I'd been running the headless CLI as a sync daemon. adb logcat pointed straight at a native `ClassCastException: Double cannot be cast to Long` inside Capacitor's `Filesystem.setTimes`: a synced document had a fractional-millisecond mtime, and the Java binding only accepts an integer `Long`.

The float gets in from the CLI. On Linux `fs.Stats.mtimeMs`/`ctimeMs` come back with sub-millisecond precision, e.g. `1778511180024.462`. The earlier fix (3f7bb047) floored these at the scan/stat adapters, but the chokidar watch adapter was missed, which is exactly the path the daemon runs on once it's up. So every live edit kept writing a float timestamp into the database, and any mobile client that replicated one would crash on its next launch.

This fixes it from both ends, as two commits:

- the CLI watch adapter now floors mtime/ctime like the other adapters, so it stops writing floats into the database in the first place;
- the Obsidian vault and storage adapters coerce mtime/ctime to integer ms at the write boundary, the last point in this repo before the native call, so a float that's already out in the mesh (from an older client, or a replica that hasn't re-synced yet) can't take the app down.

The second half is what actually un-bricked my phone: even with the remote cleaned up, the local replica still held a float and applied it on startup before it could re-sync.

### Testing

tsc-check clean and the full plugin build succeeds. CLI unit tests pass (7/7, including a new case that a fractional stat gets floored). I also verified on a real device: reproduced the launch crash from a float-mtime document, then confirmed the built plugin launches cleanly and stays up. I haven't run the CLI E2E suite locally though. 

The single right choke point is `dbToStorage` in livesync-commonlib, where the timestamp is read off the DB document once for every platform. A guard there would cover the CLI and webapp too, but it's in the `src/lib` submodule so it can't go in this PR; happy to raise it as a follow-up. Guarding per write method in the Obsidian adapters leaves a drift risk (a new write method could forget the wrapper); I considered a construction-time proxy over the write family instead, but that trades explicit type-checked calls for string-matched interception, which isn't worth it for two small adapters. On the CLI side the floor call is now hand-rolled at a few stat-read sites; a shared `fs.Stats` to `UXStat` converter would be a tidier home, but that's a wider refactor across pre-existing code.


vrtmrz:
Thank you for this fix. I tested it against v1.0.0!

The Git merge itself is conflict-free, but `npm run check` stops in the new `sanitizeWriteOptions.ts` file:

```
error TS2307: Cannot find module '@lib/common/types'
```
The repository now consumes Commonlib as an external authoritative package, so could you change that import to:
```
import type { UXDataWriteOptions } from "@vrtmrz/livesync-commonlib/compat/common/types";

```
With that one-line adjustment in an isolated integration tree, the focused CLIStorageEventManagerAdapter.unit.spec.ts test passed all 7 tests, and the complete npm run check passed. Only the
existing settings-search adoption warning remained.
Would you mind if I asked you to amend this PR, please?

vrtmrz:
Thank you again for this contribution, and I am sorry that I have left it waiting.

As it happens, Issue 1087 has now independently reproduced the same Android failure mode: a fractional timestamp was passed to Capacitor's Filesystem.setTimes, causing the native Double-to-Long cast to terminate the application.

That report made me revisit this pull request and recognise afresh that both parts of your contribution — preventing the CLI from producing fractional timestamps and defensively sanitising them at the Obsidian write boundary — are fixes that we should have in place.

Since the main branch has moved considerably and the Commonlib import paths have changed, I will merge this contribution from my side and make the small adjustments needed for the current codebase. I will also add any necessary regression coverage and rerun the current checks.

Thank you once again for identifying and fixing this so thoroughly.

PR #1055, opened 2026-07-30
Complete and correct the Korean translations

motolies:
## Summary

This supersedes vrtmrz/livesync-commonlib#79, reopened here as you requested after the v1.0.0 project-boundary change moved application translations into this repository. Thank you for the guidance, and for your work on the restructuring!

Korean coverage was 796 of 1132 keys (70%), so a large part of the UI fell back to English. This PR completes the Korean translation to 1132/1132 and fixes several user-visible defects in the existing entries.

### Fixes (carried over from the original PR)

- Three messages had lost the `$` from their placeholders, leaving a literal `{origin}`, `{org}` and `{from}->{to}` on screen, because `$msg()` only substitutes `${...}`.
- `dialog.yourLanguageAvailable` referenced `%{Display language}`, but the key is `Display Language`, so the raw token was shown. Corrected to match ja.
- Sixteen onboarding messages ended with the Japanese full stop `。`.
- Four messages hard-coded a particle after a variable (`${measuredSize}가`), which is ungrammatical for values ending in a consonant. They now use the paired form `이(가)`, `을(를)`, `(으)로`.

### Terminology

Now consistent with the Obsidian Korean UI and with the definitions in docs/terms.md: `vault` is 보관함, `device` is 기기, `status bar` is 상태 표시줄, `chunk` is 청크 everywhere, `Compaction` (압축 정리) is separated from `Data Compression` (데이터 압축), and HTTP `Origin` is 출처.

### New in this PR

The 17 keys newly introduced in v1.0.0 are translated as well: the P2P setup wizard `Invitation`, `OutroNewP2PUser` and `RebuildEverythingP2P` dialogues, and the `moduleCheckRemoteSize` notices.

### Verification

- `_tools/checkI18nCoverage.ts`: ko 1132/1132 base keys, 0 missing.
- `${...}` / `{HERE}` placeholder parity against en verified for every key; all `%{...}` cross-references resolve under the rosetta keyword rules.
- `tsc --noEmit` and a dev build pass.
- Resources were regenerated with `npm run i18n:bake` per docs/adding_translations.md. Besides the `ko` entries, the bake also trimmed a stale trailing space from five es/zh entries in `combinedMessages.prod.ts` whose YAML sources had already been trimmed during the migration; the equivalent es.json/zh.json refreshes were reverted to keep this PR scoped to Korean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EtFsro96LHztmoYmRjtfAy

PR #1075, opened 2026-08-05
Complete the Korean translations to full coverage

motolies:
## Summary

Following on from #1055, Korean coverage had fallen to 1132 of 1450 keys as the catalogue grew, so a substantial part of the UI fell back to English again. This PR completes it to 1450/1450.

All 712 namespaced keys were already translated. Every missing entry was one of the flat keys that are looked up by their English source text, which means an absent translation fell back silently rather than surfacing as an error.

### Scope

The gap was concentrated in the areas added since the previous Korean pass:

- Peer-to-Peer synchronisation UI: peers, Room ID, TURN, signalling, and broadcasting.
- Setup wizard confirmation dialogues, including the rebuild and server initialisation flows.
- S3/MinIO/R2, CouchDB, and JWT authentication settings.
- Diagnostics, maintenance, and log panels.
- A large number of short labels such as `Off`, `Sync`, `Refresh`, `Rev`, and `Stat`.

### Terminology

Consistent with the existing catalogue and docs/terms.md: `vault` is 보관함, `device` is 기기, `chunk` is 청크, `passphrase` is 패스프레이즈, `replication` is 복제, and `rebuild` is 재구축. Proper nouns and protocol names, namely CouchDB, LiveSync, Setup URI, JWT, TURN, and S3/MinIO/R2, are left as they are.

Terms settled for the first time in this PR: `broadcast` is 브로드캐스트, `signalling server` is 시그널링 서버, `Room ID` is 룸 ID, and `Group ID` is 그룹 ID.

Short labels were disambiguated by their call sites rather than by the word alone: `Rev` and `Stat` are table headings in GlobalHistory.svelte and become 리비전 and 상태, whereas `Stats` heads the P2P status card and becomes 통계.

### Concatenated fragments

Several setup wizard strings are rendered as `{fragment} <strong>{fragment}</strong>`. Korean places the predicate at the end of a sentence, so translating the trailing fragment as a bare noun phrase would leave the sentence unfinished.

In RebuildEverything.svelte the emphasised fragment, `the single, authoritative master copy`, therefore carries the predicate, and the preceding fragment ends with the subject it belongs to. The same reasoning was applied to the `As the next step,` fragments in the Outro dialogues and to the `To minimise the creation of new conflicts` fragment in FetchEverything.svelte.

### Verification

- `_tools/checkI18nCoverage.ts`: ko 1450/1450 base keys, 0 missing.
- `${...}` placeholder parity against en verified for every added key, with no mismatches. The two remaining catalogue-wide differences are pre-existing and correct: `dialog.yourLanguageAvailable` keeps the `%{Display Language}` spelling fixed in #1055, and `K.long_p2p_sync` intentionally appends `%{exp}`.
- The entries reported as identical to English are unchanged apart from `Setup-URI` and `URL`, which are deliberately kept as they are.
- `npm run check` and `npm run test:unit` pass.
- Resources were regenerated with `npm run i18n:bake` per docs/adding_translations.md. Only ko.yaml was hand-edited, and its diff is 318 additions with no deletions, so no existing translation was disturbed.

I do not have a suitable environment for the CLI End-to-End tests, so please run CI tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01R5avHoUL5zBwBfakGHyeFz


vrtmrz:
Thank you for your PR! Fine translations enrich the plug-in!
I will be happy to merge them.

PR #1083, opened 2026-08-07
Fix special characters in multi-part settings QR codes

calvinbui:
I added my new phone using the QR code method and saw that it cut off my E2EE passphrase.

- Preserve percent-encoded settings while aggregating multi-part QR codes.
- Prevent characters such as `#`, `&`, `+`, and `%` from corrupting passwords and passphrases.
- Add regression coverage for an escape sequence split across QR chunks.
- Add an Unreleased note.


vrtmrz:
Thank you for your PR! I have not realised these points!

PR #1123, opened 2026-08-20
Wire the display-language translator into the settings manifest

nimula:
After finishing the translation catalogue, I noticed the Settings pages still
did not render the translated text. I used AI to review the related code and
it raised a few suggestions below; I am not certain whether they fit the
existing design architecture, but I am including them here for reference.

## Summary
- Commonlib's `getConfig(key, translate?)` and `getConfName(key, translate?)` default `translate` to `englishMessageTranslator`, and this plug-in never passed the second argument, so every automatically wired setting rendered its name and description in English, whatever `displayLanguage` was set to.
- `settingConstants.ts` now re-exports Commonlib's names explicitly and adds thin `getConfig`/`getConfName` wrappers which default the translator to `translateLiveSyncMessage`. That reaches all three existing call sites: the 102 `setAuto`/`autoWire*` calls across the setting panes, the setup-wizard configuration summaries, and the externally-modified-setting prompt.
- `ModuleResolveMismatchedTweaks` used `confName()`, which accepts no translator, so it gains a local `localisedConfName()` instead. Switching it to `getConfName()` would have silently dropped the `statusDisplay()` suffix, replaced the empty-string fallback for an unknown key with `${key} (No info)`, and introduced `SettingInformation` as a second source.
- English output is unchanged: every catalogue key that contains a space has a value identical to the key, so translating under the default language is idempotent.
- Of the 225 distinct name/description strings across the two Commonlib manifest tables, 160 are already catalogue keys with zh-tw translations; the remaining 65 are not catalogue keys yet and pass through unchanged. Wiring the remaining literal strings in the pane files is left for a follow-up.

## Test plan
- [x] `npm run check`
- [x] `npm run test:unit`
- [x] `npm run build`
- [ ] In Obsidian, set `displayLanguage` to Chinese (Traditional); confirm automatically wired setting names/descriptions render in Chinese, then switch back to Default and confirm English is unchanged.

### DIFF
--- a/src/modules/coreFeatures/ModuleResolveMismatchedTweaks.ts
+++ b/src/modules/coreFeatures/ModuleResolveMismatchedTweaks.ts
@@ -4,7 +4,8 @@ import {
     TweakValuesShouldMatchedTemplate,
     TweakValuesTemplate,
     IncompatibleChanges,
-    confName,
+    configurationNames,
+    statusDisplay,
     type TweakValues,
     type ObsidianLiveSyncSettings,
     type RemoteDBSettings,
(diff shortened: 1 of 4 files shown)


vrtmrz:
Thank you for the PR! A few version updates have landed since this was prepared, so the exact counts in the description now differ slightly from the current codebase. This does not affect the substance of the change, which I think is very good.

Thank you for understanding and preserving the intent of the existing design. I am very pleased to merge your contribution.

PR #1129, opened 2026-08-24
fix(i18n): Spanish catalogue placeholders and missing strings

zeedif:
## Summary

Fixes for the Spanish translation catalogue found while going through the Setup Wizard, CouchDB screen, and the Config Doctor in Spanish.

- **Broken placeholder**: the language-switch notice interpolates `%{Display language}`, but the catalogue key is `Display Language` (capital L), so the token was never resolved and every language showed the literal `%{Display language}` text. Fixed the token in `es.yaml`.
- **Missing translations**: several Setup Wizard/CouchDB strings had no Spanish entry and fell back to English (the "several short steps" intro line, "Check server requirements", "Create/Connect to database and continue", the CouchDB database-name and internal-API-check notes, etc.). Added the missing keys to `en.yaml`/`es.yaml` and removed them from the provisional English-only list.
- **Untranslated interpolated value**: the Config Doctor's "activated because of ${activateReason}" message spliced in the raw English word `updated` (or the placeholder string `you wanted(Thank you)!`) directly into the Spanish sentence, since neither call site translated it before passing it along. Routed both through the catalogue.
- **Casing**: a handful of Spanish strings kept English-style Title Case (`Probar Conexión de Base de Datos`, `Privacidad y Cifrado`, ...) instead of Spanish sentence case. Also routed the hard-coded `"OK"` button through the catalogue so it renders as "Aceptar".

Scope is intentionally limited to Spanish content plus the minimal code changes needed to make two hard-coded strings translatable at all — no other locales were touched.

**Known, out-of-scope findings** (flagging for visibility, not fixed here — they live outside this repo):
- The "All optional features are disabled" / "Customisation Sync and Hidden File Sync will all be disabled..." dialog (shown after Quick Setup via a Setup URI when no remote settings exist) is entirely hard-coded English, unreachable from `obsidian-livesync`'s own translation catalogue — it's built in `@vrtmrz/livesync-commonlib`'s `Rebuilder.informOptionalFeatures()`.
- The chunk-size "Enhance chunk size" fix explanation ("With the V3 Rabin-Karp chunk splitter...") is likewise baked into `@vrtmrz/livesync-commonlib`'s Doctor rules, not this repo.

Both would need a PR against `vrtmrz/livesync-commonlib` instead.

## Test plan
- [x] `npx tsx _tools/yaml2json.ts && npx tsx _tools/bakei18n.ts` to regenerate the compiled catalogue, then formatted the generated files with prettier
- [x] `_tools/checkI18nCoverage.ts` confirms the new strings are no longer missing from the Spanish catalogue
- [x] `npx svelte-check` / `tsc --noEmit`: no new errors introduced
- [x] `npx vitest run --config vitest.config.unit.ts`: no new failures


vrtmrz:
Thank you again for this PR. While reviewing it, I found an existing oversight on the LiveSync side.

`activateReason` currently serves two purposes: it is inserted into the displayed message, but Commonlib also compares it with the exact raw value `updated` to distinguish automatic activation from a user-requested run. Therefore, translating `updated` changes behaviour in non-default languages, rather than changing only the displayed text.

Because of this, I cannot merge the current revision unchanged. The rest of the PR looks good and can be merged if the localisation of `updated` is omitted for now. Specifically, the default value in `ModuleMigration` should remain the raw `updated` value, and the new `updated` catalogue entry can be removed. The translated manually requested reason may remain.

Would you be willing to make that adjustment, or would you prefer me to make this small change while integrating the PR? I would like to consider the proper separation of the behavioural value and the displayed reason separately in Commonlib.

vrtmrz:
Thank you again for this contribution. I am very pleased that we have been able to merge your improvements. I also appreciate the careful adjustment you made after the review, particularly preserving the raw `updated` value while keeping the user-requested reason translatable. Thank you for helping to make the Spanish setup experience more complete and consistent.

PR #1130, opened 2026-08-24
feat(setup-wizard): improve E2EE dialog UX and password field

zeedif:
## Summary

Follow-up UI proposal split out of #1129 (which now covers only the Spanish translation/text fixes). This PR is UI/UX only and is language-agnostic.

- The passphrase and "Obfuscate Properties" controls in the End-to-End Encryption dialog now only appear once encryption is enabled, instead of just being disabled while still taking up space.
- The password field's show/hide control is now an icon toggle button instead of a bare checkbox.
- Translated labels can now wrap instead of being clipped into a fixed-width column — this was clipping longer translated labels (Spanish, German, ...).
- The onboarding invitation link's 44px touch-target padding is now scoped to mobile only, so on desktop it renders as a normal inline link instead of a stray, oddly-padded button.

## Test plan
- [x] `npx svelte-check` / `tsc --noEmit`: no new errors introduced
- [x] `npx vitest run --config vitest.config.unit.ts`: no new failures
- [x] Verified against the existing E2E assertions in `test/e2e-obsidian/scripts/couchdb-manual-setup-workflow.ts` (checks the encryption checkbox before interacting with the passphrase/obfuscate rows, compatible with the new conditional rendering) and `test/e2e-obsidian/scripts/onboarding-invitation.ts` (asserts the 44px minimum touch target under mobile emulation via `body.is-mobile`, which this change preserves)


vrtmrz:
Thank you for this PR. The conditional E2EE controls make the setup dialogue easier to follow, and the password visibility control and invitation-link layout are useful improvements. The translated `aria-label`, `title`, and `aria-pressed` state introduced here are also good and should be retained.

Before merging, would you be willing to make the following small adjustments?

1. Please use Obsidian's `setIcon()` with `eye` and `eye-off` rather than embedding the SVG paths directly. This keeps the control aligned with Obsidian's icon set and avoids maintaining duplicated SVG markup. If `setIcon()` does not render cleanly in this layout, an emoji is also acceptable; I have used the same pragmatic workaround elsewhere in LiveSync.

2. The change from `width: 8em` to `min-width: 8em` applies to every `InputRow` in every Svelte dialogue. I understand that this helps longer translated labels, but the global change may also alter wrapping and input widths outside the Setup Wizard. Please scope it to the affected dialogue or rows.

3. Please add focused assertions for the new behaviour. The existing CouchDB manual setup E2E workflow already reaches this dialogue, so extending it to confirm that the E2EE fields are initially hidden, appear when encryption is enabled, and retain their value while password visibility is toggled should be sufficient. A separate E2E scenario is not necessary.

Once these points are addressed, I think this will be ready to merge. Thank you for improving this workflow.

vrtmrz:
Thank you again for this contribution. The adjustments you made in response to the review have made this workflow much better. During final desktop and mobile validation, I found one minor detail: the password visibility button’s touch target was narrower than 44 pixels on mobile. I added a small style scoped to mobile and a focused regression assertion while integrating the PR, and the change has now been merged. Thank you again for improving this setup experience.

PR #1144, opened 2026-08-27
fix(storage): do not skip the initial scan when the file cache is partially warmed

### LINKED ISSUE #1143: CLI daemon deletes database entries at startup: the initial scan is skipped when the file cache is partially warmed
## Summary

On a headless CLI daemon, the startup mirror scan deleted **76 documents from CouchDB** four seconds after the container restarted. Every one of those 76 files was present on local storage at that moment, and still is.

The cause is in `NodeFileSystemAdapter.getFiles()`: it decides whether an initial directory scan is needed by looking at `fileCache.size`. Because the daemon replicates from CouchDB *before* it scans, the cache is already partially populated by `refreshFile()`, so the scan is skipped and `getFiles()` returns a **truncated** listing. The reconciliation logic then sees the remaining documents as locally deleted and removes them from the database.

Version: CLI `1.0.18`. The flawed method is unchanged on `main` (`011a840`).

## Evidence

The same file, healthy at 07:14, deleted at 09:54 — the container restarted at 09:54:42:

```
2026-08-27T07:14:10.741Z [Daemon] STORAGE == DB :Inbox/veille-cyber.md
2026-08-27T09:54:42Z      (container restarted)
2026-08-27T09:54:46.415Z [Daemon] NEWER_WINS: Treating missing local file as deletion (Inbox/veille-cyber.md)
2026-08-27T09:54:46.415Z [Daemon] DELETE DATABASE: Inbox/veille-cyber.md
...
2026-08-27T09:54:46.700Z [Daemon] Synchronisation completed: 528 files processed (98 completed, 430 skipped, 0 failed)
2026-08-27T09:54:46.700Z [Daemon] Initialized, NOW TRACKING!
2026-08-27T09:54:46.700Z [Daemon] Mirror scan complete
```

`0 failed` — the run reported success.

The file was, and remains, readable inside the container:

```
$ docker exec livesync-cli ls -la /data/Inbox
-rw-r--r-- 1 node node 10428 Aug 27 04:15 veille-cyber.md
```

76 `DELETE DATABASE` lines in that single scan.

## Root cause

**1. The scan is skipped when the cache is non-empty** — `src/apps/cli/adapters/NodeFileSystemAdapter.ts`:

```ts
async getFiles(): Promise<NodeFile[]> {
    if (this.fileCache.size === 0) {
        await this.scanDirectory();
    }
    return Array.from(this.fileCache.values());
}
```

`fileCache` has two writers: `scanDirectory()` (full, recursive) and `refreshFile(p)` (one entry). A single `refreshFile()` call is enough to make the cache non-empty and suppress the initial scan permanently.

**2. The daemon warms the cache before scanning** — `src/apps/cli/commands/runCommand.ts`:

```ts
// 1. Replicate CouchDB → local PouchDB so the mirror scan has content to work with.
const replResult = await core.services.replication.replicate(true);
// 2. Mirror scan to reconcile PouchDB ↔ local filesystem.
const scanOk = await performFullScan(core, log, errorManager, false, true);
```

Step 1 materialises documents to storage; each write goes through `refreshFile()`. By the time step 2 calls `collectFilesOnStorage()` → `storageAccess.getFiles()`, the cache holds only the handful of paths touched by replication.

**3. Missing entries are read as user deletions** — `serviceFeatures/offlineScanner`:

```js
if (options.mode === FullScanModes.NEWER_WINS && state === "db-only" && doc) {
    const lastSeenMTime = getFileMTimeFromMap(fileMapKey);
    if (lastSeenMTime !== undefined) {
        const recency = compareMTime(lastSeenMTime, doc.mtime);
        if (recency === BASE_IS_NEW || recency === EVEN) {
            action = "delete-db";
        }
    }
}
```

Every document without a local counterpart falls into `state === "db-only"`, and any of them already known to the mtime map is deleted.

That last condition explains why only a subset was hit: files written since the mtime map was last persisted are unknown to it and were spared. Two files in the same directory were treated differently — `Dispositifs/Aide à l'innovation PME.md` (older) deleted, `Dispositifs/SME Packages.md` (newer) kept.

## Impact

- **Silent data loss.** The scan logs `0 failed` and completes normally.
- **The affected files cannot recover on their own.** They are now `storage present` + `db deleted` → `both-db-deleted`. With `extraOnLocal`/`extraOnRemote` undefined (as the daemon calls it), `shouldDeleteLocalWhenRemoteDeleted()` is `false` and `extraOnLocal !== APPEND_STORAGE_ONLY`, so the pair resolves to `skip` on every subsequent scan. Re-writing the same path is refused as `db-only-deleted`; only a new path propagates.
- **It repeats on every restart**, and the blast radius grows as the mtime map learns more of the vault.
- Local files are not deleted, so nothing is lost on disk — but the vault silently stops converging.

## Suggested fix

Track scan completion explicitly rather than inferring it:

```ts
private hasScannedFully = false;

async getFiles(): Promise<NodeFile[]> {
    if (!this.hasScannedFully) {
        await this.scanDirectory();
        this.hasScannedFully = true;
    }
    return Array.from(this.fileCache.values());
}
```

`src/apps/webapp/adapters/FSAPIFileSystemAdapter.ts` carries the same pattern.

I have a PR ready with this change plus a regression test.

## A second suggestion, separate from the fix

Independently of this particular cache bug, a scan that finds **no or very few local files while the database is populated** currently proceeds to delete the database. Any future failure of the local enumeration — permissions, a mount not ready, an I/O error — would produce the same outcome through a different path.

A sanity check before mass deletion (abort, or require confirmation, when the local listing is empty or implausibly small relative to the database) would turn a silent data-loss event into a loud refusal. I'd be glad to open that separately if you think it's worth doing.

nsanitas:
Fixes #1143.

`getFiles()` inferred "the initial scan has already happened" from `fileCache.size !== 0`. The cache has a second writer, `refreshFile()`, which adds a single entry at a time — so one incidental entry was enough to suppress the initial `scanDirectory()` and make `getFiles()` return a silently truncated listing.

The CLI daemon hits this on every start, because of the order in `runCommand.ts`: it replicates from CouchDB first (materialising documents to storage, each going through `refreshFile()`), then runs the mirror scan. `collectFilesOnStorage()` therefore sees only the files touched by replication, every other document resolves to `state === "db-only"`, and `NEWER_WINS` deletes from the database those already known to the mtime map.

On the vault where this was found, that removed 76 documents four seconds after a restart, with `0 failed` in the log, while all 76 files were present on disk.

## Change

Replace the inference with explicit state, in both adapters carrying the pattern:

- `src/apps/cli/adapters/NodeFileSystemAdapter.ts`
- `src/apps/webapp/adapters/FSAPIFileSystemAdapter.ts`

```ts
private hasScannedFully = false;

async getFiles(): Promise<Files[]> {
    if (!this.hasScannedFully) {
        await this.scanDirectory();
        this.hasScannedFully = true;
    }
    return Array.from(this.fileCache.values());
}
```

The flag is set in `getFiles()` rather than inside `scanDirectory()`, since the latter recurses into subdirectories and is also callable directly with a relative path.

Behaviour is otherwise unchanged: still exactly one scan per adapter instance, and incremental updates remain the watcher's job.

## Test

New `src/apps/cli/adapters/NodeFileSystemAdapter.unit.spec.ts`, three cases against a real temporary vault:

1. a cold cache lists every file;
2. **a cache warmed by one `refreshFile()` call still lists every file** — this is the regression;
3. the scan runs only once across repeated calls.

Case 2 fails on the current implementation:

```
AssertionError: expected [ 'a.md' ] to deeply equal [ 'a.md', 'sub/b.md', 'sub/deep/c.md' ]
```

`npm run test:unit` on the touched spec: 3 passed. `npx eslint` on the touched sources: 0 errors.

## Not addressed here

A scan that finds no local files while the database is populated still proceeds to delete the database. This PR removes one way of reaching that state, not the state itself — any other failure of the local enumeration would land in the same place. I've described that in the issue and would be happy to open it separately if you'd like it handled.

### DIFF
--- a/src/apps/cli/adapters/NodeFileSystemAdapter.ts
+++ b/src/apps/cli/adapters/NodeFileSystemAdapter.ts
@@ -20,6 +20,15 @@ export class NodeFileSystemAdapter implements IFileSystemAdapter<NodeFile, NodeF
     readonly vault: NodeVaultAdapter;
 
     private fileCache = new Map<string, NodeFile>();
+    /**
+     * Whether a full directory scan has completed at least once.
+     *
+     * This must not be inferred from `fileCache.size`: the cache is also
+     * populated one entry at a time by `refreshFile()`, so a single
+     * incidental entry would suppress the initial scan and make
+     * `getFiles()` silently return a truncated listing.
+     */
+    private hasScannedFully = false;
 
     constructor(
         private basePath: string,
@@ -92,8 +101,9 @@ export class NodeFileSystemAdapter implements IFileSystemAdapter<NodeFile, NodeF
     }
 
     async getFiles(): Promise<NodeFile[]> {
-        if (this.fileCache.size === 0) {
+        if (!this.hasScannedFully) {
             await this.scanDirectory();
+            this.hasScannedFully = true;
         }
         return Array.from(this.fileCache.values());
     }
--- a/src/apps/cli/adapters/NodeFileSystemAdapter.unit.spec.ts
+++ b/src/apps/cli/adapters/NodeFileSystemAdapter.unit.spec.ts
@@ -1,0 +1,66 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { fsPromises as fs, os, path } from "@vrtmrz/livesync-commonlib/node";
+import { NodeFileSystemAdapter } from "./NodeFileSystemAdapter";
+
+describe("NodeFileSystemAdapter.getFiles", () => {
+    const tempDirs: string[] = [];
+
+    async function createVault(files: Record<string, string>) {
+        const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "livesync-cli-fs-adapter-"));
+        tempDirs.push(tempDir);
+        for (const [relativePath, content] of Object.entries(files)) {
+            const fullPath = path.join(tempDir, relativePath);
+            await fs.mkdir(path.dirname(fullPath), { recursive: true });
+            await fs.writeFile(fullPath, content, "utf-8");
+        }
+        return { adapter: new NodeFileSystemAdapter(tempDir), vaultPath: tempDir };
+    }
+
+    afterEach(async () => {
+        await Promise.all(tempDirs.splice(0).map((dir) => fs.rm(dir, { recursive: true, force: true })));
+    });
+
+    it("lists every file on a cold cache", async () => {
+        const { adapter } = await createVault({
+            "a.md": "a",
+            "sub/b.md": "b",
+            "sub/deep/c.md": "c",
+        });
+
+        const files = await adapter.getFiles();
+
+        expect(files.map((file) => file.path).sort()).toEqual(["a.md", "sub/b.md", "sub/deep/c.md"]);
+    });
+
+    it("still performs the initial scan when the cache was warmed by refreshFile", async () => {
+        // Reproduces the daemon's startup order: replication materialises a few
+        // files (each one going through refreshFile) *before* the mirror scan
+        // calls getFiles(). Inferring "already scanned" from a non-empty cache
+        // made getFiles() skip scanDirectory() and return only that subset.
+        const { adapter } = await createVault({
+            "a.md": "a",
+            "sub/b.md": "b",
+            "sub/deep/c.md": "c",
+        });
+
+        await adapter.refreshFile("a.md");
+
+        const files = await adapter.getFiles();
+
+        expect(files.map((file) => file.path).sort()).toEqual(["a.md", "sub/b.md", "sub/deep/c.md"]);
+    });
+
+    it("scans only once across repeated calls", async () => {
+        const { adapter, vaultPath } = await createVault({ "a.md": "a" });
+
+        await adapter.getFiles();
+        // A file appearing without notifying the adapter must not be picked up by
+        // a second getFiles(): the cache is authoritative once the scan has run,
+        // and incremental updates are the watcher's job.
+        await fs.writeFile(path.join(vaultPath, "b.md"), "b", "utf-8");
+
+        const files = await adapter.getFiles();
+
+        expect(files.map((file) => file.path)).toEqual(["a.md"]);
+    });
+});
--- a/src/apps/webapp/WebAppRuntime.ts
+++ b/src/apps/webapp/WebAppRuntime.ts
@@ -160,8 +160,10 @@ export class WebAppRuntime {
             throw new Error("The WebApp core is not initialised");
         }
 
+        // Discarding the cache also discards the full-scan state, so the
+        // `getFiles()` inside `collectFilesOnStorage()` rebuilds the inventory.
+        // Scanning here as well would traverse the whole Vault twice.
         fileAccess.fsapiAdapter.clearCache();
-        await fileAccess.fsapiAdapter.scanDirectory();
 
         const log = (message: unknown, level: LOG_LEVEL = LOG_LEVEL_INFO, key?: string): void => {
             this.addLog(message, level, key);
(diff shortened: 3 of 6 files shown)


vrtmrz:
Thank you for the careful diagnosis and for providing a focused regression test. I have reproduced the partially warmed-cache failure against the current `main`. Your finding around `getFiles()` is correct, and the CLI test covers the reported failure well.

Before merging this, I found three follow-on cases in the proposed scan-state handling that I think we should address:

- In `FSAPIFileSystemAdapter`, `clearCache()` does not reset `hasScannedFully`. `renameFile()` clears the cache and then repopulates only the renamed path. After an initial full scan, a rename can therefore leave `getFiles()` returning only that path. In my reproduction, the expected result contained `b.md` and `renamed.md`, but only `renamed.md` was returned.
- Both adapters currently catch directory-enumeration errors inside `scanDirectory()` and return normally. `getFiles()` then sets `hasScannedFully = true`, so a partial inventory can be accepted as complete and will not be retried. Since omitted files may subsequently be interpreted as deletions, I think a full scan should fail closed and publish its rebuilt inventory only after the complete traversal succeeds.
- The WebApp path currently performs an explicit `scanDirectory()` immediately before calling `getFiles()`. With this change, that path performs two full Vault scans.

#1145 overlaps with the Node side and contains a useful atomic, fail-closed inventory implementation. We are reviewing its daemon lifecycle ordering separately. My preferred direction would be to preserve your diagnosis and regression coverage, align the Node implementation with that stronger inventory invariant, reset the FSAPI full-scan state whenever its cache is invalidated, avoid the WebApp double scan, and add FSAPI regressions for a partially warmed cache, rename/cache invalidation, and traversal failure.

Would you be willing to update the pull request in that direction? If you would prefer, I am also happy to take over the integration work while preserving your contribution and attribution.

nsanitas:
Thank you for the review, and for reproducing all three cases.

I agree with each of them. I have addressed the two that this pull request introduced, and left the third to your integration with #1145, for the reason given below.

**Reset the FSAPI full-scan state on cache invalidation.** `clearCache()` now clears `hasScannedFully` alongside the two caches. Your `renameFile()` case is covered by a new regression, which fails without the fix with exactly the listing you observed:

```
AssertionError: expected [ 'renamed.md' ] to deeply equal [ 'b.md', 'renamed.md' ]
```

**Avoid the WebApp double scan.** There were two such sites, and the second became one only because of the fix above:

- `start()` called `scanDirectory()` immediately before `getFiles()`, as you noted.
- `scanLocalFiles()` called `clearCache()` and then `scanDirectory()`, and `collectFilesOnStorage()` calls `getFiles()`. While `clearCache()` left the scan state set, the repopulated cache suppressed that second traversal; once the state is reset, it no longer does.

Both explicit scans are removed and the traversal is left to `getFiles()`. The runtime unit tests now assert the single-traversal invariant rather than the explicit call.

**FSAPI regressions** for a partially warmed cache, for rename and cache invalidation, and for the single-scan guarantee, written against the existing `MemoryDirectoryHandle` harness in `FSAPIStorageAdapter.unit.spec.ts`.

The branch is rebased onto current `main` (1.0.24). `npm run check` passes, and `npm run test:unit` reports 857 tests passing across 113 files.

**On the fail-closed inventory, I have deliberately left it to you.** It is not a property this pull request changes: before it, a partial scan also left the cache non-empty, and suppressed the retry in the same way. It is a pre-existing weakness and I agree it should be fixed. But the right shape for it is the atomic inventory in #1145, and aligning the Node adapter with that invariant means reconciling two pull requests whose merge order and daemon lifecycle you are already reviewing. You are far better placed to do that than I am, so please do take that part over. I would rather you land the stronger invariant once than have me guess at its final form here.

I am happy to review or test whatever you land. Please run CI tests, as I cannot run the E2E suite locally.

nsanitas:
Closing this in favour of #1191, which is merged and fixes the root cause more thoroughly than
this PR did.

This PR added an explicit `hasScannedFully` flag so the initial scan would no longer be inferred
from `fileCache.size === 0`. #1191 removes the inference entirely — `getFiles()` now scans into a
fresh `Map` on every call and never consults the cache, so a partially warmed cache cannot
truncate the enumeration at all.

Thanks to @YakupEmreYerli for #1188 and to @vrtmrz for the integration and the added tests.

PR #1145, opened 2026-08-28
fix(cli): rebuild full vault inventory before mirror scan

devintucker24:
## Summary

- rebuild the CLI filesystem inventory from disk instead of treating any non-empty targeted cache as complete
- fail closed on directory/stat errors and publish the new cache only after a complete successful traversal
- share concurrent full scans and retry when targeted cache mutations overlap a traversal
- preserve rooted-path and symbolic-link safety checks
- add adapter and daemon-startup regressions to the default CLI unit suite

## Problem

The CLI daemon replicates CouchDB before running its startup mirror scan. A targeted remote reflection can populate `NodeFileSystemAdapter.fileCache` with only the changed file. Previously, `getFiles()` skipped its recursive scan whenever that cache was non-empty, so the scanner received an incomplete storage inventory.

With `NEWER_WINS` and persisted last-seen metadata, physically present files omitted from that inventory can be classified as locally missing and lead to unrelated `DELETE DATABASE` actions.

A second safety problem was that the recursive scan caught I/O errors and returned a partial cache as success. Concurrent `getFiles()` calls could also clear and populate the shared map non-atomically.

## Fix

`getFiles()` now:

1. shares one complete scan between concurrent callers;
2. builds into a private inventory;
3. uses filesystem operations that propagate ordinary I/O failures;
4. keeps the old cache unless the traversal completes successfully;
5. retries if a targeted cache mutation overlaps the traversal;
6. fails closed after repeated overlapping mutations;
7. atomically replaces the cache only after a complete stable scan.

The scan continues to validate rooted paths and refuses to traverse symbolic links.

## Regression coverage

- partial cache pre-populated by one targeted lookup still returns every disk file
- a failed subtree scan rejects instead of returning a partial inventory
- concurrent `getFiles()` callers share one atomic traversal
- explicit scans cannot escape the vault root
- symbolic links outside the vault are not traversed
- daemon ordering (`replicate` before `performFullScan`) sees the complete disk inventory
- adapter and daemon test files are included in `npm run test:unit`

## Verification

- `npm run test:unit` — 96 tests passed
- `npm run check` — passed
- `npm run build` — passed
- repeated disposable two-peer CouchDB drills covered remote-before-startup delivery, bidirectional continuous updates, intentional deletion, and divergent same-file restart/conflict behavior without unrelated deletions

### DIFF
--- a/src/apps/cli/adapters/NodeFileSystemAdapter.ts
+++ b/src/apps/cli/adapters/NodeFileSystemAdapter.ts
@@ -3,10 +3,9 @@ import type { IFileSystemAdapter } from "@vrtmrz/livesync-commonlib/compat/servi
 import { NodePathAdapter } from "./NodePathAdapter";
 import { NodeTypeGuardAdapter } from "./NodeTypeGuardAdapter";
 import { NodeConversionAdapter } from "./NodeConversionAdapter";
-import { NodeStorageAdapter } from "@vrtmrz/livesync-commonlib/node";
+import { NodeStorageAdapter, fsPromises, path, validateStoragePath } from "@vrtmrz/livesync-commonlib/node";
 import { NodeVaultAdapter } from "./NodeVaultAdapter";
 import type { NodeFile, NodeFolder, NodeStat } from "./NodeTypes";
-import { path } from "@vrtmrz/livesync-commonlib/node";
 import type { CliDiagnosticReporter } from "@/apps/cli/cliOutput";
 
 /**
@@ -20,6 +19,9 @@ export class NodeFileSystemAdapter implements IFileSystemAdapter<NodeFile, NodeF
     readonly vault: NodeVaultAdapter;
 
     private fileCache = new Map<string, NodeFile>();
+    private fileCacheVersion = 0;
+    private completeScan: Promise<NodeFile[]> | undefined;
+    private reportedScanErrors = new WeakSet<object>();
 
     constructor(
         private basePath: string,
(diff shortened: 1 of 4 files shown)


vrtmrz:
Thank you for tracking this down. I think your finding around `NodeFileSystemAdapter.getFiles()` is correct: a non-empty targeted cache must not be treated as a complete vault inventory, and scan failures should fail closed.

While investigating the startup path more closely, I found what looks like a second issue at the CLI daemon lifecycle level. In Obsidian, `useOfflineScanner(core)` binds `vault.scanVault()` to `performFullScan()`. The common lifecycle invokes this scan during `onReady()`, before replication is resumed.

The CLI uses the same lifecycle, but it does not currently compose `useOfflineScanner`, so the `scanVault()` call succeeds as a no-op. The daemon then compensates by explicitly running remote replication before `performFullScan()`. This reverses the normal Obsidian ordering and allows targeted remote reflection to populate the cache before the first complete local scan.

I think the most coherent fix may therefore be to:

1. compose `useOfflineScanner(core)` for daemon startup only, since database-only CLI commands should not implicitly scan or mirror the filesystem;
2. treat a failed `onReady()` or startup scan as a daemon startup failure;
3. perform the daemon’s initial remote replication after that local scan, and remove the post-replication scan ordering; and
4. retain your complete, fail-closed `getFiles()` work, because that adapter invariant remains valid independently of the startup ordering.

Would you be willing to investigate this direction and update the PR? If you would prefer, I am also happy to take over the lifecycle part on our side while preserving your adapter work.

PR #1152, opened 2026-08-30
Add Minimal status style: hide when idle, pill while synchronising, dismissible error

### LINKED ISSUE #233: How can I hide livesync status from bottom right bar?
Screenshot 2023-06-29 at 14 49 43

Am i able to hide or move it?

NelsonScott:
## Summary

Adds a **Status style** option under *General Settings → Appearance* with two values:

- **Classic (always visible)** — the default; the existing behaviour, unchanged.
- **Minimal (only while synchronising or on error)** — nothing is shown while the replicator is idle (`PAUSED`, `CLOSED`, `COMPLETED`, `NOT_CONNECTED` with no queued work). While a batch is moving, a small pill with a spinner and the transfer counts (`↑2 ↓1`) is shown inside the editor and stays for 1.5 s after the last transfer. When the replicator reports `ERRORED`, a `Sync failed` pill with `Retry` and a dismiss button stays until it is dismissed. Tapping the pill opens the log. The status bar item gets matching compact states (check mark / spinner with counts / warning).

On phones the pill is placed in the header row, centred between the floating buttons, so it never covers the text of the note. On desktop it keeps the current position (top right, below the header).

## Motivation

The always-on `Sync: 💤 ↑ 8 ↓ 3` line is the most visible piece of the plug-in and it is on screen 99 % of the time, when there is nothing to report. On iOS it is pinned below the header and rides over the note text while scrolling. This has been requested for a while (#233, open since 2023, "show only when syncing"). Making it an opt-in style keeps the current behaviour for everybody who relies on it.

## Implementation notes

- `src/modules/features/StatusStyle.ts`: the style constants, the `statusStyle` accessor, `deriveMinimalState()` (replicator status + a sum of every queue and in-flight counter), and `formatMinimalCounters()`. Unit tests in `StatusStyle.unit.spec.ts`.
- `ModuleLog.ts`: `renderMinimalStatus()` renders the pill and the compact status bar item; the classic path is untouched. The activity state is lingered with the existing `createMinimumVisibleActivityCount()`. The style is re-applied on `EVENT_SETTING_SAVED`, so switching it does not need a restart.
- `PaneGeneral.ts`: the dropdown. **The `statusStyle` key is not yet part of the shared settings type in Commonlib**, so it is read and written through `Reflect` and the control is rendered directly rather than through the setting spec. It is persisted in `data.json` like every other setting. I am happy to send the matching Commonlib change (type, default and `configurationNames` entry) and then move the control onto the spec if you would prefer that.
- `styles.css`: pill and status bar styles; uses the theme variables only (`--background-primary`, `--shadow-s`, `--text-error`, `--view-header-top-offset`, ...). Icons are Lucide via `setIcon`.
- `docs/settings.md` and `en.yaml` (one new message, `Sync failed`; the settings labels).

## Testing

- `npm run check`, `npm run test:unit` (724 tests) pass. Before the change, a build from the unmodified checkout reproduced the 1.0.21 release `main.js` byte for byte, so the diff in the built plug-in is only this change.
- Manually verified on Obsidian 1.13.7 desktop (Linux) and iOS with a CouchDB remote in LiveSync mode: idle shows nothing; editing a note shows the pill with `↑1` for the burst and then hides it; the status bar mirrors it. The error state has only been verified visually by class toggling, not against a real `ERRORED` replicator.
- E2E tests were not run locally. Please run CI tests.

Closes #233 if you consider this an acceptable answer to it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PybAiU9mV8Xas9GytwCGBD


NelsonScott:
@vrtmrz This is the opt-in "show status only while synchronising" idea from #233, kept behind a setting so the default is unchanged. Happy to split the `statusStyle` key out into a Commonlib PR first if you would rather have it on the shared settings type. Please run CI tests.

vrtmrz:
Thank you for putting this together. An opt-in minimal status is a useful response to #233, and preserving the existing Classic presentation as the default is the right compatibility choice.

I applied the pull request to the current `main` and reviewed the settings route, replication interaction, and accessible controls. The core presentation idea is sound, but I found several integration points that need to be addressed before we can merge it.

Most importantly, I would like this setting to use the declarative settings architecture rather than adding another bespoke imperative control to `PaneGeneral`.

`GeneralSettingSpecs` is already the shared source used by both the Obsidian 1.13 declarative settings definitions and the legacy renderer. Once `statusStyle` is a typed setting, this should be a contained change: add its dropdown specification to `createGeneralSettingSpecGroups()` and remove the manual `renderStatusStyleSetting()` implementation. This will make the setting available consistently on both settings paths and keep rendering and persistence within the established architecture.

The current implementation only adds the control to `PaneGeneral`. On the current `main`, the standard Obsidian 1.13 Appearance page is generated from `GeneralSettingSpecs`, so `statusStyle` is absent from that declarative catalogue. This appears inconsistent with the manual verification described in the pull request. Please add a focused regression test confirming that the setting is present in both the declarative and legacy paths.

To support this, `statusStyle` also needs to become part of the authoritative Commonlib settings model, including its type, default value, and configuration metadata. The current `Reflect` access bypasses that contract. As you suggested, would you be willing to open the matching Commonlib pull request first? Once it has been released, this pull request can update the pinned dependency, use the typed setting directly, and remove the `Reflect` helpers.

I also found the following points:

- The Retry control currently calls the legacy `replicate(true)` entry point. As this is an explicit user action, it should use `replicateUserInitiated()` and declare the intended interaction and progress-presentation policy, following the current manual replication actions.
- Retry and Dismiss are clickable `div` elements, while the surrounding pill is also mouse-clickable. These need semantic, keyboard-operable controls with accessible names, without creating nested interactive controls. The mobile targets are currently 24–32 px high, so they should also meet the 44 px touch-target requirement.
- The active icon uses an infinite CSS rotation. Please provide a `prefers-reduced-motion` treatment so that users who request reduced motion do not receive a continuously rotating indicator.
- The error presentation has only been checked by toggling CSS classes. Please add coverage for the actual `ERRORED` state, runtime style changes, Retry, Dismiss, and the resulting replication request.

If keeping this manageable would help, I think it would be reasonable to split the work after the typed, declarative setting and the basic idle/active presentation. The error, Retry, and Dismiss interaction could follow separately once its accessibility and runtime coverage are ready. The typed, declarative setting should remain part of the base change because it is the entry point for the feature.

Would you be willing to revise the pull request along those lines? I would be happy to review the Commonlib change first.

PR #1188, opened 2026-09-13
fix(cli): always perform the initial directory scan in getFiles (#1143)

### LINKED ISSUE #1143: CLI daemon deletes database entries at startup: the initial scan is skipped when the file cache is partially warmed

YakupEmreYerli:
Fixes #1143.

`NodeFileSystemAdapter.getFiles()` decided whether to scan the vault by checking `fileCache.size === 0`. `refreshFile()` also writes to that cache, and the daemon replicates before its mirror scan, so a single refreshed file was enough to skip the initial scan. `getFiles()` then returned a truncated listing and the reconciliation treated the missing files as locally deleted.

This tracks the full scan with an explicit flag, set once the top-level `scanDirectory()` completes. Behaviour for a cold cache and for files deleted after the scan is unchanged.

Tests: added `NodeFileSystemAdapter.unit.spec.ts`. The first case fails on `1.0.28` (returns only the refreshed file) and passes with this change. `npm run test:unit` in `src/apps/cli`: 89 passed. `npm run build` succeeds.

Branched from the `1.0.28` tag; `getFiles()` is identical on `main`.


vrtmrz:
Thank you, @YakupEmreYerli, for the fix and regression tests. They helped reproduce and address the incomplete file listing.

I had also overlooked a problem with the startup scan order itself: the daemon was replicating before scanning the local Vault.

I have incorporated an adapted version of your fix in #1191. `getFiles()` now enumerates the current filesystem independently of the cache, and the daemon and mirror scans run during database initialisation, following the Obsidian startup sequence. The daemon completes its scan before replication. Your contribution is also acknowledged in `updates.md` via #1192.

The regression tests and CLI E2E tests pass, and the fix is available in `ghcr.io/vrtmrz/livesync-cli:edge`. Would you be willing to pull the latest image and try it, particularly the daemon restart behaviour that this PR addresses?

YakupEmreYerli:
Thank you, @vrtmrz, for adapting the fix, for catching the startup order, and for the credit in updates.md.

I tested the daemon restart behaviour. I built the CLI from the `1.0.29` tag, which includes #1191, and ran the same scenario against `1.0.28` for comparison:

- Daemon A uploads a vault of 500 notes. Daemon B mirrors it.
- A is stopped. While it is down, a third local database edits 30 of those notes and syncs them.
- A is restarted, so its startup replication has 30 incoming changes.

| | 1.0.28 | 1.0.29 |
| --- | --- | --- |
| `DELETE DATABASE` lines | 155 | 0 |
| Notes left in the remote database / vault B | 345 of 500 | 500 of 500 |
| Remote edits on A's disk | yes | yes |

The 1.0.29 result held on two runs. The same test with only 20 notes and one edit passed on 1.0.28 too. The bug needs the replicated writes to reach the cache before the scan lists the directory, so a larger vault makes it reliable to reproduce.

A small note for anyone who scripts against the logs: `[Daemon] Mirror scan complete` is no longer printed in 1.0.29, since the scan now runs before `Initial replication complete`. My first test run waited for that line and timed out.

The Docker image builds cleanly from the tag with the unchanged `src/apps/cli/Dockerfile`. I'm moving my server from my patched fork to 1.0.29 now.

Closing this PR, since #1191 supersedes it. Thanks again!

PR #1190, opened 2026-09-14
fix(cli): cap exponential backoff exponent to prevent Infinity in polling loop

nxn94:
## Summary

Fix a silent-tight-loop bug in the CLI daemon's polling-mode exponential backoff: \`Math.pow(2, n)\` returns \`Infinity\` once \`n\` exceeds ~53 (since \`2^54 > Number.MAX_SAFE_INTEGER\`), which propagates through the downstream \`Math.min(...)\` into \`setTimeout(fn, Infinity)\`. Node clamps that to **1ms**, producing a non-yielding polling loop after enough consecutive failures instead of the intended capped backoff.

## The fix

A 1-line guard in \`src/apps/cli/commands/runCommand.ts\` (line 139):

\`\`\`diff
- currentIntervalMs = Math.min(baseIntervalMs * Math.pow(2, consecutiveFailures), maxIntervalMs);
+ // Cap the exponent so Math.pow(2, n) cannot overflow to Infinity (would
+ // propagate into setTimeout(fn, Infinity) → 1ms → tight loop). 30 is well
+ // within Number.MAX_SAFE_INTEGER; the downstream Math.min(..., maxIntervalMs)
+ // still bounds the final interval to the configured cap.
+ const cappedExponent = Math.min(consecutiveFailures, 30);
+ currentIntervalMs = Math.min(baseIntervalMs * Math.pow(2, cappedExponent), maxIntervalMs);
\`\`\`

## Regression test

\`daemonCommand.unit.spec.ts\` — new test **"polling backoff: consecutiveFailures overflow does not produce Infinity interval"**:

- Simulates 100 consecutive failed polls
- Asserts every scheduled \`setTimeout\` delay is finite, positive, and ≤ the configured cap (300000 ms / 5 min)
- Asserts > 54 polls were processed (proving the test exercises well past the overflow boundary)

## Verification

- Existing polling backoff test still passes (interval escalation + recovery halving behavior unchanged)
- All 13 tests in \`daemonCommand.unit.spec.ts\` pass
- \`npm run build --workspace self-hosted-livesync-cli\` succeeds

## Other bugs found in the same investigation (NOT fixed by this PR — out of scope)

While diagnosing why the polling daemon appeared to never tick in our test setup, I found two additional bugs in compiled code that lives **outside this repo's source tree** (either in the bundled \`@vrtmrz/livesync-commonlib\` or in \`@smithy/core\`):

1. **Rate limiter using a future timestamp** (\`dist/index.cjs:7654\`):
   \`\`\`js
   this.previousReplicated = now + REPLICATION_ON_EVENT_FORECASTED_TIME;  // should be \`= now\`
   \`\`\`
   Setting the timestamp to the **future** means the next event-triggered replication computes \`elapsed = now - this.previousReplicated\` which is **negative**, and \`if (elapsed < least)\` is always true → the rate limiter blocks every subsequent event-triggered replication after the first.

2. **Negative timeout in retry handler** (\`dist/extensions-DkDzTVCO.cjs:3398\`):
   \`\`\`js
   await new Promise((r) => setTimeout(r, retryDelay));
   \`\`\`
   \`retryDelay\` from the backoff math can be negative (no \`> 0\` guard). Node's \`TimeoutNegativeWarning: -X is a negative number. Timeout duration was set to 1.\` fires, and the retry tight-loops instead of backing off.

Both of these are candidate follow-up PRs against the relevant upstream packages.


vrtmrz:
Thank you for investigating this and adding the repeated-failure test.

I ran the test file from `40693659` against the unchanged implementation at `b3d01598`. All 13 tests passed without the exponent cap. I also expanded the added test to exercise 1,100 consecutive failures, and it still passed with finite, positive delays within the five-minute cap.

There are two issues with the proposed explanation:

- `Math.pow(2, 54)` is finite. `Number.MAX_SAFE_INTEGER` concerns integer precision; `Math.pow(2, n)` first becomes `Infinity` at `n = 1024`.
- `Math.min(Infinity, 300000)` returns `300000`, so the existing cap prevents that intermediate value from reaching `setTimeout`.

These unit tests cover the mocked polling behaviour with `--interval 30`. They do not reproduce the separate runtime symptoms described in the PR.

Would you be willing to clarify the observed behaviour and share a minimal reproduction, including:

- Whether polling stopped, requests repeated rapidly, or both, and when this happened relative to startup.
- The LiveSync revision, Node.js version, remote backend, and CLI command, including the `--interval` value.
- Startup and failure logs with timestamps, plus the first timeout warning and its stack trace. Node's [`--trace-warnings`](https://nodejs.org/api/cli.html#--trace-warnings) option can provide the warning stack.
- Any local source or bundle changes, and whether applying only the exponent cap changes the behaviour after rebuilding with the same dependencies.

As a precaution, please remove any credentials or other sensitive information before sharing.

This should help distinguish the daemon's polling timer from retries inside a dependency. The added test may be useful coverage, but it currently passes without the proposed fix. The rate-limiter and negative retry-delay reports can be investigated separately once their source packages and versions are identified.

nxn94:
Thank you for the careful review — you're right, and I was wrong on the premise.

I conflated `Number.MAX_SAFE_INTEGER` (≈ 2^53-1, a *precision* boundary) with the floating-point *magnitude* overflow threshold (≈ 2^1024, where `Number.MAX_VALUE` sits). `Math.pow(2, 54)` is finite (~1.8e16), and `Math.min(Infinity, 300000)` does correctly clamp even if the exponent ever reached the overflow range. Your empirical confirmation that the unmodified code passes the same test (and your expanded 1,100-failure variant) is conclusive — the bug I claimed exists doesn't.

Closing this PR as no-change. Apologies for the noise.

If you'd like, I can dig into the other two issues separately (rate limiter writing a future timestamp into `previousReplicated` in the bundled commonlib, and the unguarded `setTimeout(r, retryDelay)` in `@smithy/core`'s retry handler) — just say the word and I'll open focused issues with proper source-package identification.

